Issue #1, opened 2015-11-17
Result IDs start at 0 in SPIR-V binary produced by llvm-spirv


pierremoreau:
After compiling the [attached OpenCL kernel](https://phabricator.pmoreau.org/P66) using `clang -cc1 -emit-llvm-bc -triple spir-unknown-unknown -cl-spir-compile-options "" -include /usr/local/include/opencl_spir.h -o hello_word.bc ~/softwares/compute_test/hello_word.cl`, I ran `llvm-spirv` on the [generated llvm code](https://phabricator.pmoreau.org/F12243), which produced the [attached `hello_world.spv`](https://phabricator.pmoreau.org/F12244) file. Feeding it to `spirv-dis` from the SPIRV-Tools repository resulted in the error "error: 15: Error: Result Id is 0".
The result id of OpExtInstImport is indeed 0 in the SPIR-V binary, which is invalid by the SPIR-V spec, as IDs should be the interval `0 < id < max_bound`.

pierremoreau:
Ok, I had a look at the code and found the issue in `SPIRVModule.cpp`. `LLVMToSPIRV::transBuiltinSet()` calls `SPIRVModuleImpl::importBuiltinSet()` which sets `ExtSetId`. `SPIRVModuleImpl::importBuiltinSet()` in turns calls `SPIRVModuleImp::getId()` without any arguments, so `Id == SPIRVID_INVALID` and `Increment == 1`. As `Id` is an invalid id, `getId()` will instead return `NextId`, which was initialised to 0.

The following changes would be needed:
- Initialise `NextId` to 1;
- `isValid(SPIRVId Id)` should also return false if `Id == 0`.

I'll open a pull request with those changes later today.

PR #2, opened 2015-11-19
[SPIRV] Have SPIRV IDs start at 1

pierremoreau:
The SPIR-V specification states that an ID `id` is valid if
`0 < id < max_bound`. Therefore, an id of 0 is invalid and will be now
correctly reported by `isValid(SPIRVId)`.

Fixes: KhronosGroup/SPIRV-LLVM#1


dneto0:
Yes, I agree this is a real bug.  Also, the fix looks ok, but I haven't worked in this project before.

Issue #3, opened 2015-11-19
Generated Linkage Attribute OpDecorate doesn't contain a literal string


pierremoreau:
Translating [this LLVM code](https://phabricator.pmoreau.org/F12263) using `llvm-spirv` results in [this SPIR-V binary](https://phabricator.pmoreau.org/F12264), where the `OpDecorate LinkageAttributes` only has the linkage type but no name. (The corresponding binary stream is `4700 0400 0500 0000 2900 0000 0000 0000`.)

The code assumes the linkage type is the first attribute, which seems wrong when looking at the specification. Furthermore, there is no way to set both attributes as only the `SPIRVDecorate*` constructor modifies the `Literals` vector storing the attributes, and it only takes one word.

Note: If needed, the initial OpenCL kernel can be found [here](https://gist.github.com/pierremoreau/586f3e5ed1b9688dbf88).

yxsamliu:
"the OpDecorate LinkageAttributes only has the linkage type but no name"

By 'name' do you mean the function name?

pierremoreau:
Could be the function name maybe, I am not really sure what it is supposed to be. From the specification:

> Within a module, imported or exported values are decorated using the Linkage Attributes Decoration. This decoration assigns the following linkage attributes to decorated values:
> -    A Linkage Type.
> -    A name, which is a Literal String, and is used to uniquely identify exported values.

yxsamliu:
I see.

It is the name of the function or global variable. We should have a null terminated string (0 padded at the end if the string length plus 1 is not multiple of 4), then followed by the linkage type.

pierremoreau:
Right, that's my understanding as well. I could try to write the patch for it, but I'm unsure on how to proceed: changing the constructor to take a vector of word rather than just one word, add a setter or an `addLiteral()`/`appendWord()`, or another approach? Or if you have an idea on how to proceed, feel free to do it. :-)

yxsamliu:
I am thinking to derive SPIRVDecorateLinkageAttr from SPIRVDecorate. Add a constructor to accept name and linkage type, and pipe them into the SPIRVDecorateGeneric::Literals. Also add a convenience member function to get the name as std::string and linkage type. This way it is still a normal SPIRVDecorate and has minimal impact on other codes.

Have utility functions to convert a string to a vector of word and back may be useful for future use and make the code clearer.

I am working on some other bugs. Your help is really appreciate it. Otherwise I will fix it. :-)

[...]

PR #4, opened 2015-11-21
Add support for OpDecorate LinkageAttributes name

pierremoreau:
This still needs some work, but at least it works for my use case (see #3) and I'll be able to get some feedback on the patch.

I'm thinking for the v2 to add a patch changing the `SPIRVDecorate` constructor to take multiple words, in order to support all decorations taking a literal number, as that number might span on multiple words. This new constructor could then be used by the `OpDecorateLinkageAttr` for passing the name and type of the linkage, and avoid doing `WordCount += getSizeInWords(Name) + 1u;`.
I should also run `clang-format` for the v2, but I have to check how it works first, since I never used it.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVDecorate.h
+++ b/lib/SPIRV/libSPIRV/SPIRVDecorate.h
@@ -120,6 +120,28 @@ public:
   void setWordCount(SPIRVWord);
 };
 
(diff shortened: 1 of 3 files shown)


yxsamliu:
LGTM. Thanks.

PR #7, opened 2015-11-25
Fix SPV_VERSION definition

pierremoreau:
The specification defines the version number as

> 0 | Major Number | Minor Number | 0

which results in version `1.00` being defined as `0x00010000`. The previous
value was set to `10000`, resulting in `0x00002710`.

See: KhronosGroup/SPIRV-Tools#15

### DIFF
--- a/lib/SPIRV/libSPIRV/spirv.hpp
+++ b/lib/SPIRV/libSPIRV/spirv.hpp
@@ -46,7 +46,7 @@ namespace spv {
 
 typedef unsigned int Id;
 
-#define SPV_VERSION 10000
+#define SPV_VERSION 0x10000
 #define SPV_REVISION 2
 
 static const unsigned int MagicNumber = 0x07230203;
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -61,7 +61,7 @@ namespace SPIRV{
   /// The LLVM/SPIR-V translator version used to fill the lower 16 bits of the
   /// generator's magic number in the generated SPIR-V module.
   /// This number should be bumped up whenever the generated SPIR-V changes.
-  const static unsigned short kTranslatorVer = 4;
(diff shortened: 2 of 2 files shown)


lib/SPIRV/libSPIRV/spirv.hpp
```diff
@@ -46,7 +46,7 @@ namespace spv {
 
 typedef unsigned int Id;
 
-#define SPV_VERSION 10000
+#define SPV_VERSION 0x10000
 #define SPV_REVISION 2
```
pierremoreau, line 50: Couldn't `SPV_REVISION` and `Revision` be removed? They do not seem to be used anywhere in the code, and SPIRV-Tools did remove it in KhronosGroup/SPIRV-Tools#9.

dneto0:
The fix is to pull in updated SPIR-V headers from the SPIR-V registry.
This issue has been fixed there.

(Actually, I suspect the overall right direction to go in is for Khronos to publish the SPIR-V headers as their own GitHub repo so these other projects can stop copying the stale headers into their own trees.)

pierremoreau:
Oh, you're right! I'll open a new pull request to pull the updated headers.

(:+1: for the repository with the headers)

lib/SPIRV/libSPIRV/spirv.hpp
johnkslang, line 50: I had customer request to supply macros so they could #ifdef their code.  Seemed reasonable enough, as normal C++ if()-else won't work across versions with spelling changes.

pierremoreau, line 50: I hadn't realise at that time that this was the `spirv.hpp` from the SPIR-V registry.
It would be nice to have a repository for SPIR-V headers, as it would make it easier to update them and find if they have been updated.

pierremoreau:
Reopening this pull request as updating `spirv.hpp` isn't as straight forward as expected (see #8).

PR #8, opened 2015-11-25
Update spirv.h to latest available version

pierremoreau:
The updated version fixes the definition of `SPV_VERSION` which was missing an
`0x` prefix. This resulted in the version being defined as `0x00002710`, which
was then parsed as `0.39` rather than `1.00`.
It also removes enums being marked as _internal usage only_.


yxsamliu:
For now we cannot remove the enums for internal usage since the translator uses them.
Also please remove the trailing spaces.

lib/SPIRV/libSPIRV/spirv.hpp
```diff
@@ -71,21 +71,18 @@ enum ExecutionModel {
     ExecutionModelFragment = 4,
     ExecutionModelGLCompute = 5,
     ExecutionModelKernel = 6,
-    ExecutionModelCount, /* internal use only */
```
dneto0, line 74: I've no idea if the copy in SPIR-V LLVM has been modified this way.  ?
If so, does anything break with these gone?

pierremoreau:
… I was too hasty… Didn't thought the code would use the _internal use only_ enums. I'll get some sleep, give it a proper try and be sure to test as much as I can before trying another pull request. Sorry for the noise.

lib/SPIRV/libSPIRV/spirv.hpp
yxsamliu, line 74: Well we should be able to fix this.

pierremoreau, line 74: It is only used in SPIRVEnum.h:

```
inline bool
isValid(SPIRVExecutionModelKind E) {
    return (unsigned)E < (unsigned)ExecutionModelCount;
}
```

Should I do something like:

```
inline bool
isValid(SPIRVExecutionModelKind E) {
    return (unsigned)E <= (unsigned)ExecutionModelKernel;
}
```

But this is quite messy as, if another `ExecutionModel` is later added, it will be really difficult to find all the above hacks in the code to update them. A (better) solution would be to redefine `ExecutionModelCount` locally to `ExecutionModelKernel + 1`, but this is hacky as well. What would your advice be?

yxsamliu, line 74: Well if we can have ExecutionModelCount officially defined in the spirv.hpp the problem is solved. :-) I am wondering how the spirv-tools check the validity of ExecutionModel.

yxsamliu, line 74: I opened an issue against spirv header and see if such *Count enum values can be added. Let's wait and see.

yxsamliu, line 74: I got response that adding *Count at the end of enum definition is not desirable since an extension may add other enum values which renders *Count not a generally useful solution.

I think that makes sense. Looking back at this implementation

```
inline bool
isValid(SPIRVExecutionModelKind E) {
    return (unsigned)E <= (unsigned)ExecutionModelKernel;
}
```

It may not be as bad as we thought, because this _is_ what we currently know valid enum values. Let's say if some new enum values are added and we don't know about them, chances are that our translator is not able to handle them, so these newly added values _are_ invalid values for us. So we will get assertion, and we look at them and handle them.

So I think we should be OK to remove the internally used *Count enums and change isValid() to the new implementation.

dneto0, line 74: What you could do is define your own internal header file with your own Count definitions, and then write a test that will fail once the SPV_VERSION from spirv.h is updated.  That forces you to update the Counts if they have changed.

dneto0:
@yxsamliu asked how SPIR-V Tools checks enum validity.  It does a table lookup on the enum value. 
There is an internal table for each enum describing various properties.

Note that a range check is not sufficient for opcodes at least.  There are gaps in the instruction numbering.

lib/SPIRV/libSPIRV/spirv.hpp
pierremoreau, line 74: Right, I prefer that option as it seems cleaner than why I proposed.

pierremoreau:
I'll have a look at the code in SPIR-V Tools.

dneto0:
Your best bet is to use the JSON definitions file from the SPIR-V registry.
https://www.khronos.org/registry/spir-v/api/1.0/spirv.json
Use that at build time to build whatever tables are required for SPIR-LLVM to do what it needs.

lib/SPIRV/libSPIRV/spirv.hpp
yxsamliu, line 74: I took this approach before, i.e. defining my own enum types matching the enum values in spirv.hpp. Later I switched to use enum types in spirv.hpp since there is lots of maintenance burden due to changes in spirv.hpp. Things may be easier now since SPIR-V spec is finalized. 

Using a script to generate our own enum definition based on JSON may be a good idea, though it will take some efforts.

Another approach, instead of generate home-brew enum types, we use script to generate isValid() function based on JSON, and the function will be like:

```
bool isValid(Enum E) {
  switch(E) {
    case Enum1:
    case Enum2:
    //... all defined enum values
       return true;
    default:
       return false
 }
```

or the script could check if all enum values start from 0 and are continues and it can generate a simpler isValid() function like

```
bool isValid(Enum E) {
  return (unsigned) E <= (unsigned) EnumN;
}
```

where EnumN is the largest enum value found by the script.

yxsamliu:
I am thinking probably we should checkin the fix for the SPIR-V version first and leave the internal enum values for later. I guess it will take some time for a proper fix.

pierremoreau:
I reopened the other pull request which only contained the SPIR-V version fix. I'll update this one to remove the version fix if the other pull request gets merged, and I'll work on fixing this one.

yxsamliu:
I wrote a script to automatically generate isValid() function based on spirv.hpp. I plan to clean up the old isValid() functions and replace them with the generated ones. This will help remove some internal enum values.

pierremoreau:
I had completely stopped following this one. Did you merged your scripts you were talking about, @yxsamliu ?

yxsamliu:
I created a pull request #69 to checkin the script.

pierremoreau:
Thanks! I'll have a look during the week-end on how to use it and update this pull request.

pierremoreau:
Ok, I had a look during the week-end, but ran into some issues:
- Should the new `isValid()` functions be inlined? The old versions were, but the new ones aren't and generate multiple definitions of themselves as the header is included. Inlining those (sometimes) huge switch statement is maybe not the best idea, so I could define a `SPIRVEntry.cpp` to host the definitions.
- Some of the enums that were removed, are used as default initializer, error value, etc. it should be possible to remove them by throwing an exception instead, or having an additional error flag or similar. Or should I modify the script to generate invalid values for each enum?

yxsamliu:
For the first question, I think we can add inline keyword to these functions and put them in a new file SPIRVIsValidEnum.h. Compiler will decide whether to really inline them.

For the second question, I think we need to choose other initializers case by case. LLVM does not support exception. I hope we can avoid defining the "invalid" enum value for each enum type.

pierremoreau:
One additional question: do we want `isValid()` function on mask values? There is one `isValidSPIRVMemSemanticsMask()`, but it isn't used. The current `gen_spirv.bash` doesn't handle generating such functions yet, but I modified it to do so. Should I include those functions?

pierremoreau:
Ok, here is an updated version that transitions to the new header version, except for `OpForward` and `LinkageTypeInternal` as they are "explcitly" used in the code.
`SPIRV/simple.ll` is failing though, going to have a look.

pierremoreau:
Every test succeeds now. `SPIRV/simple.ll` was failing due to not finding `OpExecutionMode` in the `OpCodeNameMap`, since I had removed it due to removing its default constructor. I readded it, but used `ExecutionModeInvocations` instead of `ExecutionModeCount`, since the latter was removed. Most lkely not the best fix, but I'm open to suggestions.

yxsamliu:
LGTM. Thanks a lot for this work!

Alexey, can you take a look too? Thanks.

lib/SPIRV/libSPIRV/SPIRVIsValidEnum.h
```diff
@@ -0,0 +1,961 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH
+// THE SOFTWARE.
+//
+//===----------------------------------------------------------------------===//
+/// \file
+///
+/// This file defines SPIR-V isValid enums.
```
bader, line 36: Minor comment:

Could we also add a note here that this file is auto-generated by tools/spirv-tool/gen_spirv.bash and it's not supposed to be edited manually?

The rest of the changes looks good to me too. Thanks!

pierremoreau, line 36: I thought about it, but since the outer layers (putting in namespace, adding the license header) was not done by the script, and I completely forgot about adding such a message, it ended up without it. I'll modify the script to generate the whole file, and add a warning message as well.

pierremoreau:
So we are keeping `OpForward` and `OpLinkageInternal` for now, and will remove them at some later point?

yxsamliu:
For now we need to keep them. Probably we will remove them later.

lib/SPIRV/libSPIRV/SPIRVIsValidEnum.h
```diff
@@ -0,0 +1,961 @@
...
+    case BuiltInWorkgroupSize:
+    case BuiltInWorkgroupId:
+    case BuiltInLocalInvocationId:
+    case BuiltInGlobalInvocationId:
+    case BuiltInLocalInvocationIndex:
+    case BuiltInWorkDim:
+    case BuiltInGlobalSize:
+    case BuiltInEnqueuedWorkgroupSize:
+    case BuiltInGlobalOffset:
+    case BuiltInGlobalLinearId:
```
bsochack: Missing WorkgroupLinearId

pierremoreau:
Sorry for the delay, here is a new version. Among the changes:
- Add a notice message in the generated header files;
- Modify the script to generate the whole header file directly, rather than only its content;
- Update the name maps;
- Update the required capabilities maps, and add some additional ones that where missing.

bader:
LGTM. Thanks.

lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
```diff
@@ -0,0 +1,484 @@
...
+  add(DecorationFPFastMathMode, "FPFastMathMode");
+  add(DecorationLinkageAttributes, "LinkageAttributes");
+  add(DecorationNoContraction, "NoContraction");
+  add(DecorationInputAttachmentIndex, "InputAttachmentIndex");
+  add(DecorationAlignment, "Alignment");
+}
+SPIRV_DEF_NAMEMAP(Decoration, SPIRVDecorationNameMap)
+
+template<> inline void
+SPIRVMap<BuiltIn, std::string>::init() {
```
bsochack, line 350: These built-in names should have "BuiltIn" prefix per this document: https://github.com/KhronosGroup/SPIRV-LLVM/blob/khronos/spirv-3.6.1/docs/SPIRVRepresentationInLLVM.rst#id9

bader, line 350: I suggest fixing it in a separate commit as it's not related to updating spirv.h to the latest available version.

Issue #6, opened 2015-11-21
get_local_id(X) marked as LinkageAttributes export


pierremoreau:
When generating a [SPIR-V binary](https://phabricator.pmoreau.org/F12264) from the sames files as in #3 ([LLVM code](https://phabricator.pmoreau.org/F12263) and [OpenCL kernel](https://gist.github.com/pierremoreau/586f3e5ed1b9688dbf88)), `get_local_id(X)` is marked as an exported function, which seems wrong as the kernel doesn't define but rather import it.

pierremoreau:
The specification defines import as:

> A declaration of a global variable or a function that exists in another module.

which could fit in this case, though I'm unsure if we can defne `get_local_id(X)` as being defined in another module, and export is defined as:

> Accessible by other modules as well.

which, I find, is quite vague.

yxsamliu:
Agree that the global variables for LocalInvocationId etc should have import linkage, since multiple modules are supposed to share the same set of global variables defined somewhere else.

yxsamliu:
This seems to be a special case of a more generic issue. All declarations should have Import Linkage in SPIR-V. There will be a fix for this and other linkage issues.

pierremoreau:
Great! I tried to have a look at it, thinking I would change the code to have at least built-in variables marked as Import, but I got stuck on trying to figure out where the `GlobalValue::AvailableExternallyLinkage` was coming from.

bader:
#29 is merged, so this issue should be resolved

pierremoreau:
Right, I had forgotten to fo it.

Issue #5, opened 2015-11-21
Capabilities required by OpDecorate are not added to SPIR-V


pierremoreau:
When generating a [SPIR-V binary](https://phabricator.pmoreau.org/F12264) from the sames files as in #3 ([LLVM code](https://phabricator.pmoreau.org/F12263) and [OpenCL kernel](https://gist.github.com/pierremoreau/586f3e5ed1b9688dbf88)), no `OpCapability Kernel` is generated (required by `OpEntryPoint Kernel`, `OpMemoryModel OpenCL`, `OpDecorate FuncParamAttr`, `OpDecorate Constant`), nor `OpCapability Linkage`, which is required by `OpDecorate LinkageAttributes`

pierremoreau:
Hum… I do realise that missing capabilities is not restricted to `OpDecorate` instructions, but is something more widespread.

yxsamliu:
Agree we should add OpCapability Kernel for those situations.

yxsamliu:
Can you create a pull request for your fix about Capability dependencies? Thanks.

pierremoreau:
I created #18.

PR #11, opened 2015-12-01
[SPIRV] github issue#5 Capabilities required by OpDecorate are not ad…

yxsamliu:
…ded to SPIR-V.

When generating a SPIR-V binary from the sames files as in #3 (LLVM code and OpenCL kernel), noOpCapability Kernel is generated (required by OpEntryPoint Kernel, OpMemoryModel OpenCL, OpDecorate FuncParamAttr, OpDecorate Constant), nor OpCapability Linkage, which is required by OpDecorate LinkageAttributes.

Add capability when adding entry points to module.
Add required capability for decorations.
Refactor getCapability() so that empty capability can be returned.
Removed CapabilityNone.
Add name map for Capability.
Rename isValid(SPIRVId) to isValidId for clarity since SPIRVId is just typedef of uint32_t.
Reorder name maps and isValid functions in SPIRVEnum.h for further auto generation.
Add lit test for CapabilityKernel and CapabilityLinkage.

### DIFF
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -1027,7 +1027,7 @@ OCL20ToSPIRV::visitCallVecLoadStore(CallInst* CI,
   if (DemangledName.find(kOCLBuiltinName::VLoadPrefix) == 0 &&
       DemangledName != kOCLBuiltinName::VLoadHalf) {
     SPIRVWord Width = getVecLoadWidth(DemangledName);
-    SPIRVDBG(bildbgs() << "[visitCallVecLoadStore] DemangledName: " <<
+    SPIRVDBG(spvdbgs() << "[visitCallVecLoadStore] DemangledName: " <<
         DemangledName << " Width: " << Width << '\n');
     PreOps.push_back(Width);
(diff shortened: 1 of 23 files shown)


pierremoreau:
I do get `OpCapabilityLinkage` and `OpCapabilityAddresses` added for the exampled linked in #5! I also get Shader, apparently required by the use of `OpDecorateBuiltin`. However, `OpCapabilityShader` itself requires `OpCapabilityMatrix`, which was not added.

yxsamliu:
OpDecorateBuiltin should not requires CapabilityShader since Kernel also uses it. This seems to be a spec issue.

pierremoreau:
I also found it weird… incoming patch to handle capabilities depending on another capability.

pierremoreau:
https://gist.github.com/pierremoreau/5ebe7c4439249aaa9a35 should take care of capabilities dependencies. Could you add it to this pull request please?

Oh, and minor nitpick: the related open issue is #5, not #6 (in your commit message). :-)

yxsamliu:
Thanks for reminding me about the commit message. I have fixed it.
I think it is better to create another pull request for your fix. I have some other changes incoming for fixing the incorrect capability Shader and Float64.

PR #18, opened 2015-12-02
[SPIRV] Fix capabilities dependencies

pierremoreau:
Capabilities might depend on other capabilities, so make sure to add those
before adding the initial capability.


yxsamliu:
LGTM. Thanks.

PR #21, opened 2015-12-03
Fix vector convert

yxsamliu:


yxsamliu:
Sorry I forgot to remove it. Now it is removed.

yxsamliu:
I did not run the lit tests. I agree we should do that.

bader:
Thanks!

PR #26, opened 2015-12-09
Fix SPIR-V generator failure to generate OpAtomicCompareExchange and some LIT tests

AlexeySotkin:
Fix SPIR-V generator failure to generate OpAtomicCompareExchange
Removed assert which checks "Op" prefix for instruction names.
Factor out transWorkItemBuiltinsToVariables function to OCLUtil namespace
Added transWorkItemBuiltinsToVariables to OCL21ToSPIRV::runOnModule()
Fix SPIRV LIT tests

### DIFF
--- a/lib/SPIRV/OCL20To12.cpp
+++ b/lib/SPIRV/OCL20To12.cpp
@@ -80,7 +80,7 @@ char OCL20To12::ID = 0;
 bool
 OCL20To12::runOnModule(Module& Module) {
   M = &Module;
-  if (getOCLVersion(M) >= 20)
+  if (getOCLVersion(M) >= kOCLVer::CL20)
     return false;
 
   Ctx = &M->getContext();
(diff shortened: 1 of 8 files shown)


yxsamliu:
Also, with your change in LLVM to SPIRV translation, shouldn't corresponding change in SPIRV to LLVM translation be made? Will the affected OCL atomic builtin functions translated back to LLVM correctly?

bader:
I suggest adding lit tests that cover these cases to avoid concerns.

AlexeySotkin:
I think translation from SPIR-V back to LLVM will not be done correctly. You right, we have to do additional changes for that. 
To unblock merging of the request I will remove atomic_compare_exchange part for now and submit it later in separate pull request.

AlexeySotkin:
Remaining part is not needed, because the tests it is intended for are deleted. Closing the pull request.

PR #27, opened 2015-12-09
Fix issue22

yxsamliu:


lib/SPIRV/OCLTypeToSPIRV.cpp
```diff
@@ -0,0 +1,303 @@
...
+OCLTypeToSPIRV::runOnModule(Module& Module) {
+  DEBUG(dbgs() << "Enter OCLTypeToSPIRV:\n");
+  M = &Module;
+  Ctx = &M->getContext();
+  auto Src = getSPIRVSource(&Module);
+  if (std::get<0>(Src) != spv::SourceLanguageOpenCL_C)
+    return false;
+
+  CLVer = std::get<1>(Src);
+  if (CLVer > kOCLVer::CL20)
```
bader, line 80: Do you think this check is necessary?
In my opinion check for source language is good enough. If we have another version of C language I would expect to see the same LLVM types for the next version too.

yxsamliu, line 80: Right. Will remove the check for version.

PR #29, opened 2015-12-14
Changing linkage type mapping

AlexeySotkin:
Linkage type mapping between LLVM and SPIRV is as follows:
llvm private/internal linkage -> SPIRV Internal -> llvm internal linkage
llvm global value definitions with available_externally linkage or with common linkage -> SPIRV Import -> llvm available_externally linkage
llvm global value declaration(w/o available_externally or common linkage ) -> SPIRV Import -> llvm external linkage
other llvm global values  -> SPIRV Export -> llvm external linkage

Print readable names for linkage names and linkage types in text mode 

Bumping up translator version


pierremoreau:
So this should fix #6 I guess: I'll give it a try later today.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -725,7 +725,7 @@ LLVMToSPIRV::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
   if (auto GV = dyn_cast<GlobalVariable>(V)) {
     auto BVar = static_cast<SPIRVVariable *>(BM->addVariable(
         transType(GV->getType()), GV->isConstant(),
-        SPIRSPIRVLinkageTypeMap::map(GV->getLinkage()),
+        transLinkageType(static_cast<GlobalValue*>(GV)),
```
yxsamliu: GV is of type GlobalVariable, do we still need this cast?

lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -149,6 +150,11 @@ class SPIRVDecorateLinkageAttr:public SPIRVDecorate{
   SPIRVLinkageTypeKind getLinkageType() const {
     return (SPIRVLinkageTypeKind)Literals.back();
   }
+
+  void encode(std::ostream &O)const {
+    getEncoder(O) << Target << Dec << getLinkageName() << SPIRVTextOnly(" ")
```
yxsamliu: should only do this if SPIRVUseTextFormat is true, otherwise still need output binary format.

Also, need corresponding decode() function.

pierremoreau:
It does fix #6, great!

lib/SPIRV/libSPIRV/SPIRVDecorate.h
AlexeySotkin: Fixed

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: No. Fixed.

yxsamliu:
sorry for the delay. Since github did not notify new commits in pull request, I just saw your last commit. It looks good to me. However it seems there are conflicts need to be resolved before it can be merged, probably due to last merge.

PR #32, opened 2015-12-14
Clang integration.

bader:
- Added SPIRVWriterPass.
- Switched from std::ostream to llvm::raw_ostream.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVDebug.h
+++ b/lib/SPIRV/libSPIRV/SPIRVDebug.h
@@ -40,8 +40,10 @@
 #ifndef SPIRVDEBUG_HPP_
 #define SPIRVDEBUG_HPP_
 
-#include "llvm/Support/raw_ostream.h"
+#include "SPIRVUtil.h"
+#ifdef _SPIRV_LLVM_API
 #include "llvm/Support/Debug.h"
+#endif
 #include <iostream>
 
 namespace SPIRV{
@@ -61,7 +63,13 @@ extern bool SPIRVDbgErrorMsgIncludesSourceInfo;
 extern bool SPIRVDbgAssertOnError;
 
 // Output stream for SPIRV debug information.
-inline llvm::raw_ostream& spvdbgs() { return llvm::dbgs(); }
+inline spv_ostream &spvdbgs() {
+#ifdef _SPIRV_LLVM_API
+  return llvm::dbgs();
+#else
+  return std::cerr;
+#endif
+}
 
 #else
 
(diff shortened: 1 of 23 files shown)


lib/SPIRV/libSPIRV/SPIRVDebug.h
```diff
@@ -40,6 +40,8 @@
 #ifndef SPIRVDEBUG_HPP_
 #define SPIRVDEBUG_HPP_
 
+#include "llvm/Support/raw_ostream.h"
```
yxsamliu: libSPIRV is designed to be LLVM independent so that it may be used by other applications.

How about abstract ostream by a macro spv_ostream and use conditional macro to define spv_ostream as std::ostream or llvm::raw_ostream ?

The same with spvdbgs().

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -388,11 +388,11 @@ SPIRVEntry::setLinkageType(SPIRVLinkageTypeKind LT) {
...
 }
 
-std::ostream &
-operator<<(std::ostream &O, const SPIRVEntry &E) {
+llvm::raw_ostream &
+operator<<(llvm::raw_ostream &O, const SPIRVEntry &E) {
   E.validate();
   E.encodeAll(O);
-  O << SPIRVNL;
+  SPIRVNL(O);
```
yxsamliu: SPIRVNL is defined as a ostream manipulator. spv_ostream can support the same syntax by defining a << operator accepting a manipulator, e.g.

typedef spv_ostream& (*spv_ostream_manip_t)(spv_ostream &IS);

spv_ostream& operator<<(spv_ostream& O, spv_ostream_manip_t M) {
  M(O);
  return O;
}

bader: I implemented that solution, but it doesn't work.
I need to dig into it more deeply, but it's clear that due to overloading rules this function is not called. llvm::raw_ostream has another candidate, which C++ compiler likes better:

raw_ostream &raw_ostream::operator<<(const void *P) {
  *this << '0' << 'x';

  return write_hex((uintptr_t) P);
}

as you can see - it doesn't work for SPIR-V conversion.

yxsamliu: looks like the easy solution is just redefine SPIRVNL as a class and define an operator<< for it.

bader: Do you want me to create a local variable of SPIRVNL type each time we need a new line in SPIRV text or create one global variable and re-use it?

yxsamliu: A local var is fine. Since the code is so simple, compiler should be able to eliminate the local var.

bader:
Sam, are you OK with the latest fixes?

yxsamliu:
I just noticed github did not send notification when new commits are added to pull request. It only send notification when new comments are added. That's the reason for my delayed response.

PR #33, opened 2015-12-15
Fix translation of get_image_width in the reader.

AndreyE:
One lost bugfix for OpImageQuerySize


yxsamliu:
It fixes the issue. Thanks!

PR #35, opened 2015-12-16
[SPIRV] Fix assertion due to duplicate operands in opencl version met…

yxsamliu:
…adata. This happens when two SPIR modules are linked together.

Remove check for OCL version in OCLTypeToSPIRV. Only check if language is C or C++.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -139,26 +139,26 @@ decodeOCLVer(unsigned Ver) {
   return std::make_tuple(Major, Minor, Rev);
 }
 
-unsigned getOCLVersion(Module *M) {
+unsigned getOCLVersion(Module *M, bool AllowMulti) {
   NamedMDNode *NamedMD = M->getNamedMetadata(kSPIR2MD::OCLVer);
   if (!NamedMD)
     return 0;
   assert (NamedMD->getNumOperands() > 0 && "Invalid SPIR");
+  if (!AllowMulti && NamedMD->getNumOperands() != 1)
+    report_fatal_error("Multiple OCL version metadata not allowed");
+
   // If the module was linked with another module, there may be multiple
   // operands.
-  int Major = 0;
-  int Minor = 0;
-  for (unsigned I = 0, E = NamedMD->getNumOperands(); I != E; ++I) {
-    MDNode *MD = NamedMD->getOperand(I);
-    if (Major == 0) {
-      Major = getMDOperandAsInt(MD, 0);
-      Minor = getMDOperandAsInt(MD, 1);
-    } else if (Major != getMDOperandAsInt(MD, 0)
-        || Minor != getMDOperandAsInt(MD, 1)) {
+  auto getVer = [=](unsigned I) {
+    auto MD = NamedMD->getOperand(I);
+    return std::make_pair(getMDOperandAsInt(MD, 0), getMDOperandAsInt(MD, 1));
+  };
+  auto Ver = getVer(0);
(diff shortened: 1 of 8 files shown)


lib/SPIRV/OCLUtil.cpp
```diff
@@ -143,10 +143,21 @@ unsigned getOCLVersion(Module *M) {
   NamedMDNode *NamedMD = M->getNamedMetadata(kSPIR2MD::OCLVer);
   if (!NamedMD)
     return 0;
-  assert (NamedMD->getNumOperands() == 1 && "Invalid SPIR");
-  MDNode *MD = NamedMD->getOperand(0);
-  unsigned Major = getMDOperandAsInt(MD, 0);
-  unsigned Minor = getMDOperandAsInt(MD, 1);
+  assert (NamedMD->getNumOperands() > 0 && "Invalid SPIR");
```
bader, line 146: This problem affects also other named metadata nodes that must be unique across all LLVM modules: OpenCL version, SPIR version, opencl.compiler.options & opencl.compiler.ext.options (which we are dropping for SPIR-V, but llvm-link will just concatenate all options from all the modules, which is also incorrect), opencl.enable.FP CONTRACT (must be the same for all SPIR modules).

I think the best way to fix this issue to enhance regularization pass and fix metadata duplication there. What do you think?

yxsamliu, line 146: We have a pass TransOCLMD to transform SPIR 1.2/2.0 metadata to SPIR-V friendly metadata. opencl extensions and optional features are handled there. currently there is an assert that they only have one operand. I can fix that.

We don't use spir version or opencl.compiler.options, therefore no need to handle them.

In the LLVM/SPIR-V translation, this function is only used by TransOCLMD to get the correct ocl version. Later passes will use SPIR-V friendly metadata.

bader, line 146: I think it's better to make getOCLVersion function a private member function of TransOCLMD class.

yxsamliu, line 146: It is also used by OCL20To12 in SPIR-V/LLVM translation.

bader, line 146: Ok. I would prefer to have original implementation of getOCLVersion and have another function that will fix duplication of metadata. Do you think we can do that?

New implementation is doing double work by checking metadata which are supposed to be constant.
Even for the first time call it's better to be initialized with the first operand of the named metadata and after that check in the loop that all other operands have the same value.

MDNode *MD = NamedMD->getOperand(0);
unsigned Major = getMDOperandAsInt(MD, 0);
unsigned Minor = getMDOperandAsInt(MD, 1);
for (unsigned I = 1, E = NamedMD->getNumOperands(); I != E; ++I) {
    MDNode *MD = NamedMD->getOperand(I);
    if (Major != getMDOperandAsInt(MD, 0)
        || Minor != getMDOperandAsInt(MD, 1)) {
      report_fatal_error("OCL version mismatch");
    }
  }

yxsamliu, line 146: It is not so useful to fix the metadata since the fixed metadata won't be used by any passes. Instead new SPIR-V metadata with different format are generated and used by later passes. In the case of OpenCL version, the original function provides the required information. The modification is to check integrity in case multiple operands exist but the semantic of the original function does not change.

Thanks for the suggested code. I will fix that.

bader, line 146: Feel free to merge your patch as you apply suggestion above. I have no other comments.
Thanks!

PR #38, opened 2015-12-17
Update README.md

pierremoreau:
-   Add some Markdown to better distinguish code, file names and branche names from regular text.
-   Transform an unordered list to an ordered one, since it describes steps that should be followed in order.
-   Fix the indentation of one list to have it as a sublist of a list item.


yxsamliu:
Thanks!

PR #48, opened 2016-02-03
[SPIR-V] Fixed encoding of OpGroupAsyncCopy.

bader:
Memory scope was encoded as literal, but SPIR-V specification defines it as ID.


yxsamliu:
Can you add a round-trip translation test for OpGroupAsyncCopy? Thanks.

bader:
Hi Sam, 

I've added a round-trip translation test. Please, take a look. 

Thanks.

PR #49, opened 2016-02-04
Fix for get_image_array_size().

scottp101:
processing for get_image_array_size() was falling into the ExtInst case and yielding:

error: 233: Invalid extended instruction number: 138

From the disassembler.

Also, since it returns a size_t, I had to add the return type to the name since, for example, in 64-bit you could have i32 get_image_width() => OpImageQuerySizeLod and i64 get_image_array_size() => OpImageQuerySizeLod which were conflicting.

### DIFF
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -43,6 +43,7 @@
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/IR/InstVisitor.h"
 #include "llvm/IR/Instructions.h"
+#include "llvm/IR/Instruction.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/Verifier.h"
 #include "llvm/Pass.h"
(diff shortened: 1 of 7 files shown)


lib/SPIRV/OCL20ToSPIRV.cpp
```diff
@@ -855,14 +856,16 @@ OCL20ToSPIRV::visitCallGetImageSize(CallInst* CI,
       assert(IsImg);
       Desc = map<SPIRVTypeImageDescriptor>(TyName.str());
       Dim = getImageDimension(Desc.Dim) + Desc.Arrayed;
-      Ret = Type::getInt32Ty(*Ctx);
+      Ret = CI->getCalledFunction()->getReturnType();
       if (Dim > 1)
         Ret = VectorType::get(Ret, Dim);
       if (Desc.Dim == DimBuffer)
-        return getSPIRVFuncName(OpImageQuerySize);
+          return getSPIRVFuncName(OpImageQuerySize,
```
yxsamliu: Can we define a function
std::string getSPIRVFuncName(Op OC, Type *RetTy);
and use that?

yxsamliu:
Does this pass the lit test OpImageQuerySize.ll?

lib/SPIRV/OCL20ToSPIRV.cpp
scottp101: Sure, Latest change has it.

scottp101:
Passing with latest change.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -363,6 +363,11 @@ getSPIRVFuncName(Op OC, StringRef PostFix) {
 }
 
 std::string
+getSPIRVFuncName(Op OC, const Type *pRetTy, bool IsSigned) {
+  return prefixSPIRVName(getName(OC) + kSPIRVPostfix::Divider + getPostfixForReturnType(pRetTy, false));
```
yxsamliu: Better floow LLVM coding standard http://llvm.org/docs/CodingStandards.html
line width should be < 80

scottp101:
Sure thing.

AndreyE:
Hi, guys,
Do those fixes follow the 'round trip' conversion agreement? It seems these commits are mostly to the writerr.
Could you fix the reader too, please?

PR #54, opened 2016-02-05
[SPIR 2.0] Add defines for address spaces of special OCL types

AndreyE:
SPIR 2.0 specification define address space for some of special
opaque types it introduces as private. See the following note:
"Note: By default, objects reside in the private address space
(number 0). No address space qualification is used to indicate
the private address space."
However ATM different vendors treat the specification differently
so to make their lifes easier this commit introduces few defines
to mitigate this.

SPIRV_OCL_SPECIAL_TYPES_DEFAULT_ADDR_SPACE  - default
SPIRV_QUEUE_T_ADDR_SPACE                    - queue_t
SPIRV_EVENT_T_ADDR_SPACE                    - event_t
SPIRV_CLK_EVENT_T_ADDR_SPACE                - clk_event_t
SPIRV_RESERVE_ID_T_ADDR_SPACE               - reserve_id_t
SPIRV_PIPE_ADDR_SPACE                       - pipe qualifed types
SPIRV_IMAGE_ADDR_SPACE                      - image types

Values what could be used for the defines above
SPIRAS_Private    - private addr. space (default)
SPIRAS_Global     - global addr. space
SPIRAS_Constant   - constant addr. space
SPIRAS_Local      - local addr. space
SPIRAS_Generic    - generic addr. space

E.g. CMake style
add_definitions(-DSPIRV_OCL_SPECIAL_TYPES_DEFAULT_ADDR_SPACE=SPIRAS_Global)
add_definitions(-DSPIRV_RESERVE_ID_T_ADDR_SPACE=SPIRAS_Private)

Notice what SPIRV_OCL_SPECIAL_TYPES_DEFAULT_ADDR_SPACE is used
as a default addr. space for queue_t, event_t, clk_event_t, and reserve_id_t
types only. SPIR 2.0 spec. if quite clear about pipe and image types which
should reside in __global addr space.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -58,9 +58,38 @@ using namespace SPIRV;
 
 namespace OCLUtil {
 
-cl::opt<enum SPIRAddressSpace>
-ReservedIdAddrSpaceForOutput("spirv-reserved-id-addr-space",
-    cl::desc("Addr space of reserved id for output"), cl::init(SPIRAS_Global));
+
+#ifndef SPIRV_OCL_SPECIAL_TYPES_DEFAULT_ADDR_SPACE
+  #define SPIRV_OCL_SPECIAL_TYPES_DEFAULT_ADDR_SPACE SPIRAS_Private
+#endif
+
+#ifndef SPIRV_QUEUE_T_ADDR_SPACE
+  #define SPIRV_QUEUE_T_ADDR_SPACE SPIRV_OCL_SPECIAL_TYPES_DEFAULT_ADDR_SPACE
+#endif
+
+#ifndef SPIRV_EVENT_T_ADDR_SPACE
+  #define SPIRV_EVENT_T_ADDR_SPACE SPIRV_OCL_SPECIAL_TYPES_DEFAULT_ADDR_SPACE
+#endif
+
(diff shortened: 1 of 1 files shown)


lib/SPIRV/OCLUtil.cpp
```diff
@@ -238,16 +255,17 @@ transVecTypeHint(MDNode* Node) {
 SPIRAddressSpace
 getOCLOpaqueTypeAddrSpace(Op OpCode) {
   switch (OpCode) {
-  case OpTypePipe:
```
yxsamliu, line 241: for completeness, we better have cases for pipe, image and sampler.

AndreyE:
According to SPIR 2.0.
sampler_t is defined as i32, so addr. space notion is not applicable.
image types and pipe qualified types are located in global memory.
So, I didn't include them here but I don't mind about images and pipes...

lib/SPIRV/OCLUtil.cpp
AndreyE, line 241: According to SPIR 2.0.
sampler_t is defined as i32, so addr. space notion is not applicable.
image types and pipe qualified types are located in global memory.
So, I didn't include them here but I don't mind about images and pipes...

AndreyE:
Sam, could you review, please? I've applied your comment for pipes and images.

Issue #36, opened 2015-12-17
Translation of OpAtomicCompareExchange is done incorrectly


AlexeySotkin:
There are 2 issues seen during translation of atomic_compare_exchange_strong and atomic_compare_exchange_weak calls to SPIR-V:
1. 5th operand of Inst is value, and its type should be scalar value not the pointer
2. AtomicCompareExchange must return the same type as Value but it returns bool
3. Operands are incorrectly ordered
Fix for the issues mentioned above were implemented in pull request #26.
Translation of OpAtomicCompareExchange[Weak] instruction to atomic_compare_exchange_[strong|weak] call has to be fixed. Argument types, argument order and different semantics of the operation in SPIR-V and OpenCL 2.0 have to be taken into account.

yxsamliu:
Thanks for opening the issue. To fix the issue both the writer and the reader paths need change.

I noticed pull request #26 was closed. Are you going to continue working on this issue? Thanks.

AlexeySotkin:
Right now I have some other tasks with high priority. I(or @AndreyE) will work on it later. If you want, feel free to fix it.

AlexeySotkin:
Should be fixed by #58

PR #58, opened 2016-02-10
Implement support of OpAtomicCompareExchange in SPIR-V generator and …

AlexeySotkin:
…consumer

Type of return value, arguments' types and thier order is corrected.


AlexeySotkin:
Should fix #36

PR #63, opened 2016-02-15
[Reader] Fix mangling of device execution built-ins with local memory arguments

AndreyE:
According to OpenCL 2.0 specification some device execution built-ins may have
ellipsis in the arguments along with a block with local memory arguments. These
built-ins apparently must be mangled differenlty then built-ins w\o blocks
with local memory arguments so BE could be able to handle them properly.

uint get_kernel_work_group_size (void (^block)(local void *, …));

uint get_kernel_preferred_ work_group_size_multiple (
 void (^block)(local void *, …));

int enqueue_kernel (queue_t queue, kernel_enqueue_flags_t flags,
                    const ndrange_t ndrange,
                    uint num_events_in_wait_list,
                    const clk_event_t *event_wait_list,
                    clk_event_t *event_ret,
                    void (^block)(local void *, ...…),
                    uint size0, ...)

Unfortuantely SPIR 2.0 specification doesn't state anything about these
built-ins except the generic sentece about built-ins mangling. So, this
commit follows SPIR 2.0 generator and SPIR 2.0 spec. and Itanium C++ ABI
spec. to mangle these built-ins.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -299,9 +299,17 @@ getOCLOpaqueTypeAddrSpace(Op OpCode) {
   }
 }
 
-class OCLBuiltinFuncMangleInfo:public SPIRV::BuiltinFuncMangleInfo {
+// Fetch type of invoke function passed to device execution built-ins
+static FunctionType *
+getBlockInvokeTy(Function * F, unsigned blockIdx) {
+    auto params = F->getFunctionType()->params();
+    PointerType * funcPtr = cast<PointerType>(params[blockIdx]);
+    return cast<FunctionType>(funcPtr->getElementType());
+}
+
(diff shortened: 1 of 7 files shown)


lib/SPIRV/OCLUtil.cpp
```diff
@@ -421,6 +447,6 @@ mutateFunctionOCL(Function *F,
 void
 llvm::MangleOpenCLBuiltin(const std::string &UniqName,
     ArrayRef<Type*> ArgTypes, std::string &MangledName) {
-  OCLUtil::OCLBuiltinFuncMangleInfo BtnInfo;
+  OCLUtil::OCLBuiltinFuncMangleInfo BtnInfo(nullptr);
```
turinevgeny, line 450: Is it consistent with assertion on line 327?

AndreyE, line 450: For the three device execution built-ins it is consistent. It uses another way to get built-ins mangled.

PR #66, opened 2016-02-19
Fix for get_image_array_size().

bader:
Add return type to overload for OpImageQuerySizeLod and OpImageQuerySize.

Refactor return mangling into function.


yxsamliu:
closed by mistake. reopen it.

PR #80, opened 2016-03-02
[Writer] Allow ndrange_2D and ndrange_3D to coexist in the same input module

AndreyE:
The commit initially made by Scot. It doesn't require fixes in the reader.

### DIFF
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -509,7 +509,9 @@ OCL20ToSPIRV::visitCallNDRange(CallInst *CI,
     default:
       assert(0 && "Invalid number of arguments");
     }
-    return getSPIRVFuncName(OpBuildNDRange);
+    // Translate ndrange_ND into differently named SPIR-V decorated functions because
+    // they have array arugments of different dimension which mangled the same way.
+    return getSPIRVFuncName(OpBuildNDRange, "_" + std::to_string(Len) + "D");
   }, &Attrs);
 }
 
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -1071,11 +1071,12 @@ Value *
 getScalarOrArray(Value *V, unsigned Size, Instruction *Pos) {
   if (!V->getType()->isPointerTy())
     return V;
-  auto GEP = dyn_cast<GetElementPtrInst>(V);
-  assert (GEP);
-  auto P = GEP->getPointerOperand();
+  assert((isa<ConstantExpr>(V) || isa<GetElementPtrInst>(V)) &&
+         "unexpected value type");
+  auto GEP = cast<User>(V);
+  assert(GEP->getNumOperands() == 3 && "must be a GEP from an array");
+  auto P = GEP->getOperand(0);
   assert(P->getType()->getPointerElementType()->getArrayNumElements() == Size);
-  assert(GEP->getNumIndices() == 2);
   auto Index0 = GEP->getOperand(1);
   assert(dyn_cast<ConstantInt>(Index0)->getZExtValue() == 0);
   auto Index1 = GEP->getOperand(2);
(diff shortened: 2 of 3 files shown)


test/SPIRV/transcoding/BuildNDRange_2.ll
```diff
@@ -0,0 +1,363 @@
...
+%opencl.block = type opaque
+%struct.ndrange_t = type <{ i32, [3 x i64], [3 x i64], [3 x i64] }>
+
+@block_fn.gs = internal constant [2 x i32] [i32 64, i32 4096], align 4
+@block_fn.gs1 = internal constant [3 x i32] [i32 64, i32 64, i32 64], align 4
+
+; Function Attrs: nounwind
+define spir_func void @block_fn(i32 %level, i32 %maxGlobalWorkSize, i32 addrspace(1)* %rnd, i32 addrspace(1)* %res) #0 {
+entry:
+  %level.addr = alloca i32, align 4
```
bader: Please, reduce the test case to have minimal number of instructions. Preferably 3 calls of built-in function with constant arguments.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1071,11 +1071,11 @@ Value *
 getScalarOrArray(Value *V, unsigned Size, Instruction *Pos) {
   if (!V->getType()->isPointerTy())
     return V;
-  auto GEP = dyn_cast<GetElementPtrInst>(V);
+  auto GEP = isa<ConstantExpr>(V) || isa<GetElementPtrInst>(V) ? cast<User>(V) : nullptr;
```
bader: Is that relevant to ndrange fix?

test/SPIRV/transcoding/BuildNDRange_2.ll
```diff
@@ -0,0 +1,363 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.spv.txt
+; RUN: FileCheck < %t.spv.txt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV-DAG: BuildNDRange {{[0-9]+}} {{[0-9]+}} [[GWS:[0-9]+]] [[LWS:[0-9]+]] [[GWO:[0-9]+]]
```
bader: What's the point of defining patterns w/o using them?

lib/SPIRV/OCL20ToSPIRV.cpp
```diff
@@ -509,7 +509,7 @@ OCL20ToSPIRV::visitCallNDRange(CallInst *CI,
     default:
       assert(0 && "Invalid number of arguments");
     }
-    return getSPIRVFuncName(OpBuildNDRange);
+    return getSPIRVFuncName(OpBuildNDRange, "_" + std::to_string(Len) + "D");
```
bader, line 514: I don't understand this change. According to description it's done to avoid some collision. There should be no collisions - all the OpenCL built-ins must map to the same "LLVM-SPIR-V" built-in __spirv_BuildNDRange_*.

AndreyE, line 514: The problem is in ndrange_2D and ndrange_3D. They are mangled the same way but have different argument types (2D and 3D arrays).

test/SPIRV/transcoding/BuildNDRange_2.ll
AndreyE: --check-prefix=CHECK-SPIRV is defined, to my understanding -DAG postfix allows the test do not relay on the order of the patterns

lib/SPIRV/SPIRVUtil.cpp
AndreyE: Yes, it is. I'm updated the test there this should be clear.

test/SPIRV/transcoding/BuildNDRange_2.ll
bader: I meant FileCheck variables GWS, LWS and GWO - they are defined, but not used.
Please, add checks for all ndrange instructions - there should be four of them and validate types of arguments.

AndreyE: Done

PR #82, opened 2016-03-02
Fix constant sampler case where the constant is still accessed through a load

AlexeySotkin:


bader:
Looks good to me.
Sam, do you have any comments?

yxsamliu:
LGTM. Thanks.

Issue #65, opened 2016-02-18
The LLVM <-> SPIRV converter is not thread safe.


romanovvlad:
For example there is undefined behavior in SPIRVMap with using static maps if converter works in parallel.

yxsamliu:
SPIRVMap is intended for static initialization only, i.e. all its instances should be statically created. If all its usages follow this rule, will it still cause thread safety issue?

AndreyE:
To answer Sam's question.
I think this should be OK because it would eliminate possible data races. But be careful with local function static variables. According to C++11 standard, which introduced multi-threading, function local variables with static lifetime is safe to initialize in multi-threading environment. But I tend to think what this might be unsupported on some compilers, maybe some older versions of VS that have partial support of C++11. So the safest way is to move local static variables to an outer scope to initialize those maps at process start or library loading.

I found the following statement in final C++11 draft:
6.7 Declaration statement
…
4 The zero-initialization (8.5) of all block-scope variables with static storage duration (3.7.1) or thread storage
duration (3.7.2) is performed before any other initialization takes place. Constant initialization (3.6.2) of a
block-scope entity with static storage duration, if applicable, is performed before its block is first entered.
An implementation is permitted to perform early initialization of other block-scope variables with static or
thread storage duration under the same conditions that an implementation is permitted to statically initialize
a variable with static or thread storage duration in namespace scope (3.6.2). Otherwise such a variable is
initialized the first time control passes through its declaration; such a variable is considered initialized upon
the completion of its initialization. If the initialization exits by throwing an exception, the initialization
is not complete, so it will be tried again the next time control enters the declaration. If control enters
the declaration concurrently while the variable is being initialized, the concurrent execution shall wait for
completion of the initialization.88 If control re-enters the declaration recursively while the variable is being
initialized, the behavior is undefined. [ Example:
int foo(int i) {
static int s = foo(2*i); // recursive call - undefined
return i+1;
}
—end example ]
…
88) The implementation must not introduce any deadlock around execution of the initializer.

romanovvlad:
We are using VS 2012 which does not support safe initialization of local static.
(https://msdn.microsoft.com/en-us/library/hh567368.aspx)
I have declared maps that initialized in getMap and getRMap function as thread_local and it fixed my problem.

bader:
Should be fixed by pull request #83.

PR #83, opened 2016-03-03
Guard initialization of SPIRVMap.

romanovvlad:
Initialization of static local variable is not thread safe unless c++11 "magic static"(MSVC) feature is supported by compiler. The msvc2012 is the minimal requirement for LLVM 3.6 and it is not support "magic static". So we should take care of safe initialization of SPIRVMap.


bader:
Fixes issue #65.

PR #85, opened 2016-03-04
Fix global block

AlexeySotkin:


AlexeySotkin:
Some non-functional i.e. white-space changes was added to lib/SPIRV/SPIRVReader.cpp accidentally.
To filter them out **?w=1** should be added to the URL. Here is the link to see my changes:
https://github.com/KhronosGroup/SPIRV-LLVM/pull/85/files?w=1

PR #87, opened 2016-03-07
Fixed comparisons between signed and unsigned integer expressions.

aaronenyeshi:
Changed to unsigned BtnInfo->getVarArg() for comparison. Added boolean BIVarArgNegative to replace BtnInfo->getVarArg() < 0 checks.


yxsamliu:
Since this change only affects mangling of OpenCL builtins with varargs and we already have a test for it (device_execution_simple_local_memory.ll), no new tests are needed.

Issue #89, opened 2016-03-09
The names of built-in variables don't match the documentation


bsochack:
The documentation (https://github.com/KhronosGroup/SPIRV-LLVM/blob/khronos/spirv-3.6.1/docs/SPIRVRepresentationInLLVM.rst#id9) says that the built-in variables should have names:
__spirv_BuiltIn{Name}

In the SPIR-V converter (https://github.com/KhronosGroup/SPIRV-LLVM/blob/khronos/spirv-3.6.1/lib/SPIRV/libSPIRV/SPIRVEnum.h#L229) these built-ins don't have "BuiltIn" prefix.
Is this generator bug or documentation?

Personally I'd prefer the names with "BuiltIn" prefix.

yxsamliu:
I think the documentation is right. The translator is done before the documentation therefore is outdated. We should fix the translator.

bsochack:
I can take care of fix. I'll open a pull request tomorrow

bsochack:
https://github.com/KhronosGroup/SPIRV-LLVM/pull/97

bader:
Fixed by #97.

PR #97, opened 2016-03-14
updated built-in variable names to match LLVM for SPIR-V documentatio…

bsochack:
…n (issue #89)

https://github.com/KhronosGroup/SPIRV-LLVM/issues/89


test/SPIRV/builtin_vars-decorate.ll
```diff
@@ -0,0 +1,78 @@
+; RUN: llvm-as < %s | llvm-spirv -spirv-text -o %t
+; RUN: FileCheck < %t %s
+
+; ModuleID = 'test.cl'
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+; CHECK: 4 Decorate [[GID:[0-9]*]] BuiltIn 24
```
bader: Here you define variable GID, which not used later, moreover it's re-defined.
I recommend either:
1. Do you not use variables: CHECK: 4 Decorate {{[0-9]+}} BuiltIn 24
2. Improve check and validate that Decoration is applied to correct variable. You can check it by validating data type of the variable.

PR #101, opened 2016-03-15
Fixed potentially uninitialized local variable Builtin used

aaronenyeshi:
Fixes error: potentially uninitialized local variable 'Builtin' used


yxsamliu:
LGTM

pierremoreau:
It will never be used unitialised, but probably still a good idea to give it a default value.

PR #102, opened 2016-03-16
Adding support for doubles with relational functions

paigeale:


lib/SPIRV/OCL20ToSPIRV.cpp
```diff
@@ -1239,14 +1239,25 @@ void OCL20ToSPIRV::visitCallRelational(CallInst *CI,
...
+            Type *ScalarTy = isDouble ? Type::getInt64Ty(*Ctx) : Type::getInt32Ty(*Ctx);
+            Type *VTy = VectorType::get(ScalarTy,
+                                        NewCI->getType()->getVectorNumElements());
           False = Constant::getNullValue(VTy);
           True = Constant::getAllOnesValue(VTy);
         } else {
-          False = getInt32(M, 0);
-          True = getInt32(M, 1);
+          if (isDouble)
+          {
```
yxsamliu: not follow LLVM code style

yxsamliu:
Please squash commits as one, and add a transcoding test.

romanovvlad:
There is a bug, we should not return int64 if arg is scalar double.
Openc C spec( rev 32 ) define following variants of isnan function, 
int isnan(float)
intn isnan(floatn)
int   isnan(double)
longn isnan(doublen)]

Also there is no fix for consumer.

I also worked on this problem and I have patch for it #105.

paigeale:
Thank you romanovvlad for fixing this for the vector case. I apologize for my mistake in implementing the scalar case I missed this when I was looking at the spec.

paigeale:
This pull request can now be disregarded in light of romanovvlad's patch. Thanks again

PR #105, opened 2016-03-17
Fix translation of relational funcs in case of arg is vector of doubles.

romanovvlad:
According to OpenCL C spec( rev 32 ) if argument of functions isnan, isnormal, isinf, isfinite is vector of doubles than return value should be longn.


bader:
LGTM.

PR #118, opened 2016-03-29
Add sampler type to SPIR-V friendly LLVM format.

yxsamliu:
The representation of sampler constant will be added later.


AlexeySotkin:
I think that OpTypeSampler should not be mapped to opaque type. Per the OpenCL spec: "sampler declared in the program source must be initialized by 32-bit unsigned integer constant..." How such initialization can be represented in LLVM IR with opaque type?
What about %opencl.sampler_t = type { i32 } ?

yxsamliu:
Marcin proposed another representation. We need to discuss that. This change is mostly for the name of the sampler type.

PR #119, opened 2016-03-29
Add SPIR-V friendly sampler representation for LLVM.

yxsamliu:


bader:
Hi Sam, the patch looks good to me, but I let AlexeyS to decide since he is currently working on exact the same problem.

yxsamliu:
This fix is mostly about changing the sampler name from opencl.sampler_t to spirv.Sampler. It does not touch the representation about the sampler constant.

lib/SPIRV/OCLTypeToSPIRV.cpp
```diff
@@ -250,8 +250,12 @@ void OCLTypeToSPIRV::adaptArgumentsBySamplerUse(Module &M) {
           AdaptedTy.count(SamplerArg) != 0) // Already traced this, move on.
         continue;
 
+      if (isSPIRVType(SamplerArg->getType(), kSPIRVTypeName::Sampler))
+        return;
```
AlexeySotkin, line 254: Maybe we should not return, but continue with rest of the users of the function, shouldn't we ?

AlexeySotkin, line 254: Except that, LGTM.

yxsamliu, line 254: We only need to adapt the i32 type sampler argument in __spirv_SampledImage. This only happens with the old SPIR 1.2/2.0 format.

If the sampler arg is already spirv.Sampler type, that means FE already generates correct type, so we do not need to adapt it.

PR #126, opened 2016-04-06
[SPIR-V] Add missing Addresses capability to SPIR-V files with non-logical addressing model.

bader:
- Bump up translator version.

### DIFF
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -62,7 +62,7 @@ namespace SPIRV{
   /// The LLVM/SPIR-V translator version used to fill the lower 16 bits of the
   /// generator's magic number in the generated SPIR-V module.
   /// This number should be bumped up whenever the generated SPIR-V changes.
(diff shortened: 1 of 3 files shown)


lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -403,6 +403,8 @@ class SPIRVTypeImage:public SPIRVType {
   SPIRVCapVec getRequiredCapability() const {
     SPIRVCapVec CV;
     CV.push_back(CapabilityImageBasic);
+    if (Desc.Dim == 1)
+      CV.push_back(CapabilitySampled1D);
```
yxsamliu: Sampled1D capability depends on Shader capability. Do we really need this?

bader: It's required by specification and checked by SPIR-V validation tool.
Dependency on Shader capability looks odd to me.
I'll revert this change while it's not cleared up.
Thanks!

PR #127, opened 2016-04-14
Accept sampler as opaque struct in LLVM IR

AlexeySotkin:


AlexeySotkin:
@yxsamliu , could you take a look.

lib/SPIRV/OCLTypeToSPIRV.cpp
```diff
@@ -1,4 +1,4 @@
-﻿//===- OCLTypeToSPIRV.cpp - Adapt types from OCL for SPIRV ------*- C++ -*-===//
+//===- OCLTypeToSPIRV.cpp - Adapt types from OCL for SPIRV ------*- C++ -*-===//
```
yxsamliu, line 1: I am wondering what change is in this line? Is it line ending?

lib/SPIRV/OCLTypeToSPIRV.cpp
```diff
@@ -294,7 +294,7 @@ OCLTypeToSPIRV::adaptArgumentsByMetadata(Function* F) {
     auto NewTy = *PI;
     if (OCLTyStr == OCL_TYPE_NAME_SAMPLER_T && !NewTy->isStructTy()) {
       addAdaptedType(Arg, getOrCreateOpaquePtrType(M,
-        getSPIRVTypeName(kSPIRVTypeName::Sampler)));
+        getSPIRVTypeName(kSPIRVTypeName::Sampler), SPIRAS_Private));
```
yxsamliu: It makes more sense to put sampler in constant addr space since they never change.

Also, we better factor out sampler type creation as a function since it is used many times.

lib/SPIRV/OCLTypeToSPIRV.cpp
AlexeySotkin, line 1: It was in UTF-8 encoding. Now it is in ASCII. First 3 bytes (byte order mark) have been removed.

yxsamliu, line 1: Thanks. I guess the file was somehow messed up before. We'd better keep all files in ASCII.

lib/SPIRV/OCLTypeToSPIRV.cpp
AlexeySotkin: Both are done

PR #128, opened 2016-04-20
[SPIR-V] Translate image access qualifiers encoded in OpenCL image ty…

bader:
…pe name.


bader:
Notes:
- This patch enables translation of images types with access qualifiers encoded in the LLVM type names. Clang part of the patch is here: https://github.com/KhronosGroup/SPIR/pull/31.
- Translation from SPIR-V to LLVM doesn't recover access qualifiers yet. Sam, do we need this part?

yxsamliu:
I think translating to the new image type name and name mangling in the SPIR-V/LLVM translation is useful, but should be put under an option which is off by default. Since vendor's builtin library is likely using the old image type names.

bader:
Agree. Depending on what clang version vendors is using to compile built-in library they may get either old names or new names. Clang changes were merged from the ToT.

I'll fix SPIR-V -> LLVM translation.

Issue #122, opened 2016-03-31
OpControlBarrier and OpMemoryBarrier to OpenCL 1/2/2.0 transformations don't work correctly


bsochack:
During validation process of https://github.com/KhronosGroup/SPIRV-LLVM/pull/121 pull request I found a few other issues:
- some arguments of OpControlBarrier are dropped during conversion from SPIR-V to OpenCL 1.2/2.0
- some arguments of OpMemoryBarrier are dropped during conversion from SPIR-V to OpenCL 1.2/2.0
- OpMemoryBarrier produces always mem_fence, but it should produce atomic_work_item_fence for OpenCL 2.0

PR #134, opened 2016-05-11
[SPIR-V] Fix translation of OpControlBarrier and OpMemoryBarrier

ghost:
Fixes: https://github.com/KhronosGroup/SPIRV-LLVM/issues/122

OpControlBarrier and OpMemoryBarrier are now correctly translated to SPIR-V and
vice versa.

OpControlBarrier will be translated to:
- barrier() if OpSource is OpenCL 1.2 and execution scope is Workgroup
- work_group_barrier() if OpSource is OpenCL 2.0 or higher and the execution
  scope is Workgroup
- sub_group_barrier() if execution scope is Subgroup

OpMemoryBarrier will be translated to:
- mem_fence() if OpSource is OpenCL 1.2
- atomic_work_item_fence() if OpSource is OpenCL 2.0 or higher

If OpSource is not specified, we assume it is the highest available version.


yxsamliu:
I think we need to check the SPIRV has correct numerical values for mem scope and mem semantics in the tests, not just format.

ghost:
@yxsamliu good point, fixed.

PR #142, opened 2016-05-20
Support for SPIRV 1.1 MaxByteOffset decoration.

ldudziak:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -653,6 +653,11 @@ LLVMToSPIRV::transFunctionDecl(Function *F) {
       BA->addAttr(FunctionParameterAttributeZext);
     if (Attrs.hasAttribute(ArgNo + 1, Attribute::SExt))
       BA->addAttr(FunctionParameterAttributeSext);
+    if (Attrs.hasAttribute(ArgNo + 1, Attribute::Dereferenceable)
+        && oclIsKernel(F))
```
yxsamliu: why limit this to kernel? this attribute can be used in non-kernel functions.

ldudziak: After double-check you are right. If I remember correctly there was such restriction at the beginning (probably because in C++ there's similar one) but I cannot see it now - I will remove this limitation.

PR #150, opened 2016-06-02
Fix bug converting SPIR-V module to text form.

bader:
llvm-spirv tool with '--to-text' option added capabilities that were not actually encoded in binary form.

Also completed fix partially implemented in pull request #135.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2603,7 +2603,6 @@ llvm::ReadSPIRV(LLVMContext &C, std::istream &IS, Module *&M,
   M = new Module("", C);
   std::unique_ptr<SPIRVModule> BM(SPIRVModule::createSPIRVModule());
 
-  BM->setAutoAddCapability(false);
   IS >> *BM;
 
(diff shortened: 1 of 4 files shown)


yxsamliu:
Can we have a test for this? translate an llvm to spirv binary first, then convert to text format and make sure it pass. Or just re-use some existing test and add an extra step?

bader:
In order to verify the fix to disassembler we need to create invalid SPIR-V module with missing capability and verify that llvm-spirv tool doesn't add non-existing capability. I'm not sure how to do that using current toolset. Are you OK, if we add a test in binary form?

yxsamliu:
How about generate spirv binary b1 first, convert to spirv text, then convert to spirv binary b2, and use

cmp -b b1 b2

to compare b1 and b2. They should be the same.

yxsamliu:
I may misunderstood. did llvm-spirv translate llvm to spirv binary missing capabilities? or it only happens with some invalid spirv binary generated by other tools?

bader:
I think the approach with testing round trip translation via SPIR-V text form will work.
I added new check to existing test.
Thanks!

PR #154, opened 2016-06-10
extractvalue/insertvalue handling

a2flo:
This adds handling of LLVM extractvalue/insertvalue instructions (e.g. as used for loading/storing a value from/to an array), and handling of the reverse translation from SPIR-V to LLVM in OpCompositeExtract/OpCompositeInsert.

Several notes:
- SPIR-V classifies structures, arrays, matrices and vectors as composite types. SPIRV-LLVM currently only handles vector types in OpCompositeExtract/OpCompositeInsert.
- I have only tested this for array types right now. LLVM struct types will also need testing. Since neither OpenCL nor LLVM have a native matrix type, I'm not sure how or if matrix types should be tested.
- the ExtractValueInst/InsertValueInst handling must go before the UnaryInstruction handling (in SPIRVWriter), because they are both unary instructions.
- indices in LLVM and SPIR-V match up, so no conversion is necessary.

Also, help creating proper test cases would be appreciated. Probably have to write these by hand as I can't seem to tickle extractvalue/insertvalue instructions out of LLVM for simple example code (it usually prefers GEPs instead of these).

As a more convoluted example, I currently have this:

```
%struct.raster_params = type { %class.matrix4 }
%class.matrix4 = type { %struct.const_array }
%struct.const_array = type { [16 x float] }
[...]
  %5 = getelementptr inbounds %struct.raster_params, %struct.raster_params addrspace(2)* %params, i64 0, i32 0, i32 0, i32 0
  %.unpack.unpack = load [16 x float], [16 x float] addrspace(2)* %5, align 4
  %.unpack.unpack.fca.2.extract = extractvalue [16 x float] %.unpack.unpack, 2
```

which should get translated to (misc stuff omitted):

```
          %5 = OpTypeFloat 32
          %6 = OpTypeInt 64 0
        %169 = OpConstant %6 16
        %170 = OpTypeArray %5 %169
[...]
        %217 = OpInBoundsPtrAccessChain %216 %183 %56 %58 %58 %58
        %218 = OpLoad %170 %217 Aligned 4
        %219 = OpCompositeExtract %5 %218 2
```

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1574,13 +1574,18 @@ SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
 
   case OpCompositeExtract: {
     SPIRVCompositeExtract *CE = static_cast<SPIRVCompositeExtract *>(BV);
-    assert(CE->getComposite()->getType()->isTypeVector() && "Invalid type");
-    assert(CE->getIndices().size() == 1 && "Invalid index");
+    if (CE->getComposite()->getType()->isTypeVector()) {
+      assert(CE->getIndices().size() == 1 && "Invalid index");
+      return mapValue(
+          BV, ExtractElementInst::Create(
+                  transValue(CE->getComposite(), F, BB),
+                  ConstantInt::get(*Context, APInt(32, CE->getIndices()[0])),
+                  BV->getName(), BB));
+    }
     return mapValue(
-        BV, ExtractElementInst::Create(
+        BV, ExtractValueInst::Create(
                 transValue(CE->getComposite(), F, BB),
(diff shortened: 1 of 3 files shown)


bader:
ExtractValue/InsertValue instructions are used for atomic operations or target-specific intrinsics.
Some examples can be found in Clang lit tests. Just grep 'extractvalue' or 'insertvalue' LLVM instruction in clang/test directory and you will a few examples.
But I would suggest taking some simple LIT tests from llvm\test\SPIRV\transcoding\ directory (e.g. atomic_store.ll) and manually add instructions you want to test.

a2flo:
Did it manually now, instead of finding a simple OpenCL example.
Any remarks? Anything missing?

bader:
I'm OK with it, but I'd like Sam to review it also.

yxsamliu:
LGTM. Thanks.

PR #156, opened 2016-06-22
fix: OpCapability Kernel missing in a few cases

nizhegorodets:


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -138,8 +140,6 @@ class SPIRVModuleImpl : public SPIRVModule {
   void setAlignment(SPIRVValue *, SPIRVWord);
   void setMemoryModel(SPIRVMemoryModelKind MM) {
     MemoryModel = MM;
-    if (MemoryModel == spv::MemoryModelOpenCL)
-      addCapability(CapabilityKernel);
```
yxsamliu, line 142: It would be cleaner to keep setMemoryModel() unchanged whereas call it in the SPIRVModuleImpl constructor.

nizhegorodets:
updated

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -71,6 +71,8 @@ class SPIRVModuleImpl : public SPIRVModule {
     MemoryModel(MemoryModelOpenCL){
```
yxsamliu: can we remove this? since it is redundant.

nizhegorodets: If remove this the function addCapability() will not be called and disappears "capability kernel" in output disassembled "spirv_file"

bader: addCapability should still be called since setMemoryModel will set the same value as this line.
I agree with Sam - line #71 is redundant.

nizhegorodets:
updated

PR #157, opened 2016-06-22
Merge LLVM 3.6.2 release into SPIRV-LLVM

InBetweenNames:
This PR is to prepare for future work in porting the SPIRV-LLVM work to LLVM 3.8.  This brings the repository up to date with release_36 in the official LLVM repository.


SimonDanisch:
Great!
Is there any Roadmap for 3.8 support? Our Julia LLVM-SPIRV backend desperately needs this!
Best would be of course, to merge this into the main LLVM repository!

InBetweenNames:
Actually, an LLVM 3.8 branch is available here in my own repository:

https://github.com/InBetweenNames/SPIRV-LLVM/tree/spirv-3.8.0_r2

There are two main issues to switching to this branch however:

1) The LLVM IR changed in 3.7, making the generated SPIR actually differ between what is in the SPIR specification.  SPIR-V generation should still work, however.

2) Clang would also need to be updated to version 3.8.  This is definitely an involved task as there already exists OpenCL 1.1 support in mainline LLVM that differs between what the SPIR repository offers.  I have started some preliminary efforts to do this, but I have nothing publishable unfortunately.

In order to set things up for the Clang 3.8 merge, I suggest we first merge in the 3.6.2 changes.  The reason is that both SPIR and SPIRV-LLVM are actually behind the 3.6 stable branch of LLVM.  Keeping them both based on the release version of Clang helps with the merge, as I have observed in my own tests locally.

As for dealing with the LLVM IR change, it may make sense to state, instead, that SPIR code should use the 3.6.2-based SPIRV-LLVM and SPIR repos since they generate compliant SPIR, and for SPIR-V code the 3.8 branches could be used (with SPIR being for internal use only in generating the SPIR-V, since this SPIR would not follow the official SPIR specification).

Regardless of which option is chosen, I think integrating the 3.6.2 changes is a good first step.  Your thoughts?

yxsamliu:
I am thinking probably it is better to have a branch tracking the LLVM trunk, so I created a new branch khronos/spirv-master based on khronos/spirv-3.6.1.

How about we do all the merging in this new branch? When it reaches LLVM 3.8 we can create a new branch khronos/spirv-3.8. And we can continue merging with LLVM master in khronos/spirv-master until it reaches ToT.

InBetweenNames:
yxsamliu: Have you looked at my own 3.8 branch of SPIRV-LLVM?  That could be made to track LLVM trunk without too much effort, and would save some work.

yxsamliu:
Your branch looks good. There is no build failure and only 4 lit test failures. How about I create khronos/spirv-master and khronos/spirv-3.8 branches based on your branch?

InBetweenNames:
Sounds good to me!

yxsamliu:
Branches created. We can continue merging efforts in the new branches.

InBetweenNames:
Great!  I'll go ahead and close this PR then since the relevant functionality is now in the spirv-3.8 branch.

PR #161, opened 2016-07-13
Update GettingStarted.rst

yxsamliu:


yxsamliu:
Update GettingStarted.rst about build LLVM with CMake since we only supports building with CMake.

bader:
Hi Sam,

The patch looks good, but I don't think it's necessary.
README.md refers to http://llvm.org/docs/GettingStarted.html, which contains build instructions for current ToT using CMake (autotools are deprecated for the ToT).

PR #165, opened 2016-07-21
Add missing capability for 1D images and images with Buffer dimension…

bader:
…ality.


AlexeySotkin:
What about capabilities for other dimensionalities: Cube, Rect, SubpassData ?

bader:
These are not used by OpenCL and require Shader capability.

AlexeySotkin:
Okay, LGTM.

Issue #169, opened 2016-08-10
llvm-spirv from text to spv: OpUndef is moved from function


echuraev:
llvm-spirv tool moves OpUndef instruction from function to program scope. It is bug in the llvm-spirv tool.
Step to reproduce:
1. Convert attached spv file to binary (llvm-spirv.exe -to-binary reproducer.spt.txt)
2. Convert spv file back to text. 
3. Undef will not be in body of function.
After fixing this bug, please implement LIT test for OpUndef in function.

[reproducer.spt.txt](https://github.com/KhronosGroup/SPIRV-LLVM/files/418104/reproducer.spt.txt)

PR #170, opened 2016-08-10
Implemented possibility use OpUndef in Basic Block in SPIR-V file.

echuraev:
LIT test wasn't written because issue #169 affect to it.

I have tested it manually but I don't have time to fix this issue. I'll do it later and will add LIT test.


lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -150,7 +150,10 @@ SPIRVFunction::decodeBB(SPIRVDecoder &Decoder) {
 
     SPIRVInstruction *Inst = static_cast<SPIRVInstruction *>(Decoder.getEntry());
     assert(Inst);
-    BB->addInstruction(Inst);
+    if (Inst->getOpCode() == OpUndef)
+      Module->addUndef(Inst->getType());
```
yxsamliu: addUndef creates Undef instruction with the next available id, which may not match the id of the Undef instruction in the original SPIRV module. Also, what if two function define the same type of Undef?

AlexeySotkin: I think, we just don't need to call addUndef, because Decoder.getEntry() crates an entry and adds it to the module. Besides, in this way, id of the original Undef instructions is kept.

AlexeySotkin: If two functions define Undef of the same type, we will have these two Undef instructions in module scope along with other constants. They will be of the same type, but with different ids.
See also #169

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -149,8 +149,9 @@ SPIRVFunction::decodeBB(SPIRVDecoder &Decoder) {
     }
 
     SPIRVInstruction *Inst = static_cast<SPIRVInstruction *>(Decoder.getEntry());
-    assert(Inst);
-    BB->addInstruction(Inst);
+    assert(Inst);
+    if (Inst->getOpCode() != OpUndef)
+      BB->addInstruction(Inst);
```
yxsamliu, line 154: The undef constant is not added to the module.

yxsamliu:
need a spirv -> llvm test.

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
echuraev, line 154: Why did you think that undef constant will not be added to the module? 
As I see in the implementation of method getEntry(), it is called at this line: M.add(Entry);.

yxsamliu, line 154: Sorry, I missed that. Can you add a 'spirv -> llvm' lit test for this? Thanks.

test/SPIRV/opundef.ll
```diff
@@ -0,0 +1,37 @@
...
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64-unknown-unknown"
+
+%0 = type { i32, float }
+
+; Function Attrs: nounwind
+define spir_func void @foo() #0 {
+entry:
+  %agg1 = insertvalue %0 undef, i32 1, 0
```
slsamliu: can you add a check that undef instruction is inside a SPIR-V function? I have some doubt about that. As I remember, the translator will generate undef as constant outside of any functions. If that's the case, we may need a hand-crafted SPIR-V module. Also, better to have two functions each containing an undef instruction of the same type.

PR #179, opened 2016-09-06
Setting name of target of LinkageAttributes decoration to literal val…

AlexeySotkin:
…ue of the decoration.

Also enabling .spt files in LIT tests

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVEntry.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVEntry.cpp
@@ -253,10 +253,14 @@ SPIRVEntry::validateBuiltin(SPIRVWord TheSet, SPIRVWord Index)const {
 }
 
 void
-SPIRVEntry::addDecorate(const SPIRVDecorate *Dec){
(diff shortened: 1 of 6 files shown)


bader:
LGTM. Thanks!

yxsamliu:
Do we have a bi-way translation test for this?

AlexeySotkin:
If you mean LLVM->SPIRV->LLVM, no. OpName instructions should be removed from SPIRV module to reproduce the issue. AFAIK there is no way to produce SPIRV without OpName with llvm-spirv tool. So I used manually stripped SPIRV module.

yxsamliu:
This change affect function declaration name since it uses linkage name to overwrite the name of the function. Do we have test for the linkage name, otherwise how do we make sure the round-trip translated llvm IR still has correct function name?

AlexeySotkin:
Many tests from test/SPIRV/transcoding directory check function names after round trip translation. For example spirv-types.ll, cl-types.ll, global_block.ll.

PR #180, opened 2016-09-08
Fix error with adding capability for 1D images

echuraev:


echuraev:
LIT test was implemented here: https://github.com/KhronosGroup/SPIRV-LLVM/commit/f185af0fe5f2f2dd47628d3b62c712cc90a15dd5

Issue #174, opened 2016-08-17
llvm.fmuladd.f32 is not code generating properly


raunintc:
when a kernel uses the llvm.fmuladd intrinsic we're not generating correct code.

The generated SPIR-V is this:

```
     OpName %8 "llvm.fmuladd.f32"
     OpDecorate %8 LinkageAttributes "llvm.fmuladd.f32" Import
%8 = OpFunction %6 Pure %7
```

   %46 = OpFunctionCall %6 %8 %65 %47 %20

I think llvm-spir needs to emit a opfmul + opfadd.

See attached for an example.

[FancyJuliaSet.txt](https://github.com/KhronosGroup/SPIRV-LLVM/files/423789/FancyJuliaSet.txt)

AlexeySotkin:
Fixed by [Translation of LLVM intrinsics #185](https://github.com/KhronosGroup/SPIRV-LLVM/issues/185)

PR #185, opened 2016-09-23
Translation of LLVM intrinsics

AlexeySotkin:
An LLVM intrinsic should not be transalated to SPIRV as a function call.
Either it's semantics is represented via SPIRV instruction(s)
or an error is reported.


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1201,6 +1220,63 @@ LLVMToSPIRV::transSpcvCast(CallInst* CI, SPIRVBasicBlock *BB) {
...
+      MemoryAccess[0] |= MemoryAccessAlignedMask;
+      MemoryAccess.push_back(AlignVal);
+    }
+    if (MI->isVolatile())
+      MemoryAccess[0] |= MemoryAccessVolatileMask;
+    return MemoryAccess;
+  };
+
+  switch (II->getIntrinsicID()) {
+  case Intrinsic::fmuladd : {
```
yxsamliu, line 1236: Probably we should translate llvm.fma.\* the same way and emit a warning.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1201,6 +1220,63 @@ LLVMToSPIRV::transSpcvCast(CallInst* CI, SPIRVBasicBlock *BB) {
...
+                                        BB);
+    return BM->addBinaryInst(OpFAdd, Ty, Mul,
+                             transValue(II->getArgOperand(2), BB), BB);
+  }
+  case Intrinsic::memset : {
+    // If memset is used for zero-initialization, find a variable which is being
+    // initialized and store null constant of the same type to this variable
+    MemSetInst *MSI = cast<MemSetInst>(II);
+    ConstantInt *Val = dyn_cast<ConstantInt>(MSI->getValue());
+    if (Val && Val->isZero()) {
```
yxsamliu: This only works if the size and alignment of the pointee type matches the length and alignment argument of memset instruction. This is not always true, so need to add these two conditions.

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: We don't have immediate need for this right now, do we? Besides there is no facilities for emitting warnings. This commit is not intended to cover all llvm intrinsics, I just wanted to fix #174 and llvm.memset which revealed during testing. Probably we should submit another issue to handle llvm.fma.*

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: Fixed

lib/SPIRV/SPIRVWriter.cpp
yxsamliu: OK. Let's leave it until there is need to translate this.

PR #186, opened 2016-09-26
OpVariable in OpSpecConstantOp.

echuraev:
In one of last change of spirv spec was added following change:
"Public SPIRV-Headers issues # 12 and # 13 and Khronos SPIR-V issue # 65: Allow OpVariable as an initializer for another OpVariable instruction or the Base of an OpSpecConstantOp with an AccessChain opcode."
In this way, OpVariable can be initialize before OpSpecConstantOp. To do this, it's necessary add OpSpecConstantOp in one container with OpVariable in TopologicalSort. After this change, OpSpecConstantOp and OpVariable instructions will be add to spirv file in the right order.


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1225,15 +1225,15 @@ class TopologicalSort {
...
   typedef std::vector<SPIRVTypeForwardPointer *> SPIRVForwardPointerVec;
   typedef std::function<bool(SPIRVEntry*, SPIRVEntry*)> IdComp;
   typedef std::map<SPIRVEntry*, DFSState, IdComp> EntryStateMapTy;
 
   SPIRVTypeVec TypeIntVec;
   SPIRVConstantVector ConstIntVec;
   SPIRVTypeVec TypeVec;
-  SPIRVConstantVector ConstVec;
-  SPIRVVariableVec VariableVec;
+  SPIRVInstructionVec InstructionVec;
```
yxsamliu: The name is confusing. How about ConstAndVarVec ?

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1269,11 +1269,11 @@ class TopologicalSort {
       if (C->getType()->isTypeInt())
         ConstIntVec.push_back(C);
       else
-        ConstVec.push_back(C);
+	InstructionVec.push_back(E);
```
yxsamliu: incorrect indentation

PR #187, opened 2016-09-26
Fix for translation of OpFMod instruction

kkyzylova:
OpFMod(a, b) should be translated to copysign(frem(a, b), b)

### DIFF
--- a/test/SPIRV/fmod.ll
+++ b/test/SPIRV/fmod.ll
@@ -8,13 +8,12 @@
 ; RUN: llvm-spirv -r %t.spv -o %t.bc
 ; RUN: llvm-dis < %t.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
-; CHECK-SPIRV: 5 FMod {{[0-9]+}} {{[0-9]+}} {{[0-9]+}} {{[0-9]+}}
+; CHECK-SPIRV: 7 ExtInst {{[0-9]+}} {{[0-9]+}} {{[0-9]+}} fmod {{[0-9]+}} {{[0-9]+}}
 ; CHECK-LLVM: call spir_func float @_Z4fmodff
 ; CHECK-LLVM: declare spir_func float @_Z4fmodff(float, float)
 
-; ModuleID = 'test.bc'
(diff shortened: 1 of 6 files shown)


bader:
LGTM.

yxsamliu:
I am not sure about this.

This change translates SPIRV OpFMod(x,y) to copysign(frem(x,y),y) where frem is llvm frem instruction, and copysign is opencl builtin.

Two questions:
1. llvm frem returns the remainder of a division. The remainder has the same sign as the dividend. Why do we still need copysign?
2. frem(-2.1, -2) should get -0.1 since that's the remainder with the same sign.
   SPIRV FMod is defined as floating-point modulo operation of Operand 1 modulo Operand 2. The sign of a non-0 result comes from Operand 2.

My understanding is that modulo of -2.1 to -2 is 1.9. By forcing the sign of dividend, we should get -1.9.

Am I missing something? Or should we ask the WG for clarification about OpFMod? Thanks.

kkyzylova:
1. In this case the remainder should have the same sign as the divisor. So
2. According to Raun Krisch:

>    OpFRem(x,y) == OpenCL.ext.fmod(x,y)
>    OpFMod(x,y) == copysign( OpenCL.ext.fmod(x,y), y )

frem(-2.1, -2) should get -0.1, yes. And the result of OpFMod is -0.1 too: copysign(-0.1, -2) = -0.1.

yxsamliu:
I see. So
OpFRem <=> llvm.frem
OpenCL.ext.fmod <=> OpenCL.ext.fmod (equivalent to llvm.frem)
OpFMod => OpenCL.ext.copysign(llvm.frem)

test/SPIRV/fmod.ll
```diff
@@ -1,51 +0,0 @@
-; __kernel void fmod_kernel( float out, float in1, float in2 )
```
yxsamliu, line 1: I think we need to keep this test but modify it to be bi-way test and match expected result. I think it should be translated to extended instruction OpenCL.fmod, right?

kkyzylova: Done

PR #189, opened 2016-09-28
Support translation of OpLifetimeStart and OpLifetimeStop instructions

kkyzylova:


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1542,6 +1543,49 @@ SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+  case OpLifetimeStop: {
+    SPIRVLifetimeStop *LTStop = static_cast<SPIRVLifetimeStop*>(BV);
+    IRBuilder<> Builder(BB);
+    SPIRVWord Size = LTStop->getSize();
+    ConstantInt *S = nullptr;
+    CallInst *Stop = nullptr;
+    if (Size)
+      S = Builder.getInt64(Size);
+    auto Var = transValue(LTStop->getObject(), F, BB);
+    for (const auto &I : Var->users()) {
```
yxsamliu: Can we refactor this part to a function getBitcastForLifeStart(X) ?

If it exists, returns the bitcast, otherwise X itself. Then just have one call of Builder.CreateLifetimeEnd.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1267,6 +1267,32 @@ LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II, SPIRVBasicBlock *BB) {
       transValue(II->getOperand(2), BB),
       getMemoryAccess(cast<MemIntrinsic>(II)),
       BB);
+  case Intrinsic::lifetime_start : {
+    int64_t Size = dyn_cast<ConstantInt>(II->getOperand(0))->getSExtValue();
+    if (Size == -1)
+      Size = 0;
+    if (isa<BitCastInst>(II->getOperand(1))) {
```
yxsamliu: Need to check the size of the real object matches the size argument of lifetime.start, otherwise cannot drop the bitcast.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1267,6 +1267,32 @@ LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II, SPIRVBasicBlock *BB) {
...
+    int64_t Size = dyn_cast<ConstantInt>(II->getOperand(0))->getSExtValue();
+    if (Size == -1)
+      Size = 0;
+    if (isa<BitCastInst>(II->getOperand(1))) {
+      auto LT = BM->addLifetimeStartInst(
+          transValue(dyn_cast<BitCastInst>(II->getOperand(1))->getOperand(0), BB),
+          Size, BB);
+      auto BC = LT->getPrevious();
+      assert(BC->getOpCode() == OpBitcast && "Invalid instruction");
+      BM->eraseInstruction(BC, BB);
```
yxsamliu: need to check the only users of the bitcast are lifetime.start and/or lifetime.end, otherwise cannot drop the bitcast.

lib/SPIRV/SPIRVReader.cpp
yxsamliu: Can you add a comment about the bitcast handling when translating lifetime.end?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1267,6 +1267,32 @@ LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II, SPIRVBasicBlock *BB) {
       transValue(II->getOperand(2), BB),
       getMemoryAccess(cast<MemIntrinsic>(II)),
       BB);
+  case Intrinsic::lifetime_start : {
```
yxsamliu: Add a comment about bitcast handling.

lib/SPIRV/SPIRVReader.cpp
kkyzylova: Everything is fixed

yxsamliu:
Please resolve conflicts. Thanks.

kkyzylova:
But I don't see any conflict to resolve. I have to create new pull request

PR #195, opened 2016-10-26
Change ordering of blocks as required by SPIRV Spec

kkyzylova:
According to specification:

> The order of blocks in a function must satisfy the rule that blocks appear before all blocks they dominate.

This pass helps to avoid the cases which are contrary to SPIRV Spec.

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -42,8 +42,10 @@
 #include "SPIRVMDBuilder.h"
 #include "SPIRVMDWalker.h"
 
+#include "llvm/ADT/SCCIterator.h"
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/ADT/Triple.h"
+#include "llvm/IR/CFG.h"
 #include "llvm/IR/InstVisitor.h"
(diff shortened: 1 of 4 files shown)


yxsamliu:
Do we have a test for this?

Also, is this sufficient to always satisfy the spec? I am wondering if we need a dedicated pass to do this.

kkyzylova:
I've changed the fix and added a regression test, please take a look. Thanks.
There is no SimplifyCFG pass now.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -154,6 +156,34 @@ SPIRVRegularizeLLVM::regularize() {
         }
       }
     }
+
```
yxsamliu, line 161: Please refactor this part as a member function and add come comments about how it works.

I guess this only fixes the specific issue for switch instruction, not a generic transformation to guarantee that blocks appearing before all blocks they dominate, right? If so it is OK, just mention that in the comment and give it a proper name e.g. fixBlockOrderForSwitch.

yxsamliu:
Please update the title for this pull request for what it does. Thanks.

yxsamliu:
It caused travis-ci failure. Can you fix that? Thanks.

PR #197, opened 2016-11-14
Revert "Change ordering of blocks as required by SPIRV Spec (#195)"

kkyzylova:
This reverts commit 53688bd35ed4c279ec93b7229e3b741b7ab98b2c.
Fix does not handle some specific cases and should be reverted.


bader:
Please, include "some specific cases" to the updated fix. Thanks.

Issue #23, opened 2015-12-07
Missing `SPIRVSubTarget.h` file


pierremoreau:
While trying to compile the new `spirv-target` branch, I get the following error:

```
In file included from ../lib/Target/SPIRV/SPIRVISelDAGToDAG.cpp:44:
../lib/Target/SPIRV/SPIRVTargetMachine.h:41:10: fatal error: 'SPIRVSubTarget.h' file not found
#include "SPIRVSubTarget.h"
```

@yxsamliu Did you forget to push this file? :-)

yxsamliu:
Did you use CMake to build? It works for me on both MSVC 2013 and cygwin.
The Makefile in the directory itself may not work.

pierremoreau:
Gotcha! The file is named `SPIRVSubtarget.h` and not `SPIRVSubTarget.h`, which will of course work on Windows since it's case insensitive, but not on Linux. As only `SPIRVTargetMachine.h` is calling it `SPIRVSubTarget.h`, I guess the name was meant to be `SPIRVSubtarget.h`. I'll write a patch.

yxsamliu:
My bad :-) Thanks for fixing this.

pierremoreau:
No problem! I would say this is one of the most common kind of errors when trying to build a Windows project on Linux/Mac OS. :-D

Issue #22, opened 2015-12-07
Duplicate `OpTypeImage` definition


pierremoreau:
In [this](https://phabricator.pmoreau.org/F13641) SPIR-V binary (produced from [this](https://phabricator.pmoreau.org/F13640) `.bc`, out of [this](https://phabricator.pmoreau.org/F13639) OpenCL file), `OpTypeImage` is defined twice, with the exact same parameters (well, except for the result id):
-  `1900 0900 0700 0000 0600 0000 0100 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000`
-  `1900 0900 1500 0000 0600 0000 0100 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000`

bader:
Hi Pierre,

Please, verify that this issue is fixed by Sam in rev. b03f4776295403f7622c56897eeea746117cd144.

Thanks,
Alexey

pierremoreau:
Hi Alexey,

Just tested the latest version of `khronos/spirv-3.6.1` and the issue is fixed! Thanks for the fix @yxsamliu!

Thanks,
Pierre

Issue #20, opened 2015-12-03
SPIR-V reader translate some conversion/bitcast instructions incorrectly to OCL builtin functions


yxsamliu:
pull request #17 caused regression for the following kernel:

__kernel void test_fn( __global char2 *src, __global short *dst )
{
  int tid = get_global_id( 0 );
  short tmp = as_short( src[ tid ] );
  dst[ tid ] = tmp;
}

clang generates the following code:

  %call = tail call spir_func i64 @_Z13get_global_idj(i32 0) #2
  %sext = shl i64 %call, 32
  %idxprom = ashr exact i64 %sext, 32
  %arrayidx = getelementptr inbounds <2 x i8> addrspace(1)\* %src, i64 %idxprom
  %0 = load <2 x i8> addrspace(1)\* %arrayidx, align 2, !tbaa !9
  %astype = bitcast <2 x i8> %0 to i16
  %arrayidx2 = getelementptr inbounds i16 addrspace(1)\* %dst, i64 %idxprom
  store i16 %astype, i16 addrspace(1)\* %arrayidx2, align 2, !tbaa !12
  ret void

SPIRV-V reader translates bitcast <2 x i8> %0 to i16 incorrectly to 

call spir_func i16 @_Z13convert_shortDv2_c(<2 x i8> %1)

Another example:
__kernel void math_kernel8( __global int8\* out, __global float8\* in1, __global float8\* in2 )
{
   int i = get_global_id(0);
   out[i] = isequal( in1[i], in2[i] );
}

The LLVM/SPIR-V translator needs to generate conversion instructions for i1 x 8 -> i32 x 8 for representing isequal() as instruction in SPIR-V. SPIR-V reader should not translate it to OCL builtin function.

yxsamliu:
Let's close this for now since the commit causing regression is reverted. We can reopen this if necessary.

AndreyE:
About isequal. Boolean vectors are not defined by OpenCL C or by SPIR-1.2/2.0 specifications. Also
sext, trunc instructions are defined for scalar types only, so the reader would violate SPIR spec if it generates such instructions for vector types.
So in SPIR-V created from SPIR 1.2/20 operations producing boolean vectors should be always accompanied by casts (sext) to integer vectors. So, it seems like it should be enough to combine these two operations into a single call to OCL built-in returning such integer vector.

AndreyE:
[...]

Issue #19, opened 2015-12-03
SPIR-V generator is crashed on a simple device execution kernel.


AndreyE:
SPRI-V converter hits the following assert during conversion from the attached LLVM IR to SPIR-V
Assertion `HasVariWC && WC >= WordCount && "Invalid word count"' failed.
To reproduce:
$ llvm-as enqueue.ll.txt -o enqueue.bc && llvm-spirv enqueue.bc
Call stack [enqueue.bt.txt](https://github.com/KhronosGroup/SPIRV-LLVM/files/50754/enqueue.bt.txt)

The problematic LLVM IL [enqueue.ll.txt](https://github.com/KhronosGroup/SPIRV-LLVM/files/50750/enqueue.ll.txt) is compiled from the following OpenCL C 2.0 source code
void __kernel one(__global float \* inout) {
     _inout = cos(_inout);
}

void __kernel two(__global float \* inout) {
     enqueue_kernel(get_default_queue(), CLK_ENQUEUE_FLAGS_WAIT_KERNEL, ndrange_1D(1), ^{one(inout);});
}

yxsamliu:
This test is invalid. The LLVM/SPIR-V translator expects arguments of enqueue_kernel follows the SPIR 2.0 format. The attached LLVM IR does not conform to SPIR 2.0.

bader:
Guys, I've uploaded OpenCL front-end compiler to https://github.com/KhronosGroup/SPIR - 'spirv-1.0' branch. It supports both versions of OpenCL languages - 1.x and 2.0, but I didn't modify blocks implementation for two reasons:
1. It's not necessary for SPIR-V translation. It was only required for SPIR-2.0 specification.
2. I'd like to contribute https://github.com/KhronosGroup/SPIR - 'spirv-1.0' branch patches to clang repository (http://llvm.org/svn/llvm-project/cfe/trunk). Modifying blocks for OpenCL doesn't make sense since it's not much different from existing implementation, so I assumed it will be rejected.

I tested this new front-end with conformance tests for 1.2 and 2.0. Everything should be fine, except blocks.
I'm going to write a short readme file with instructions how to generate SPIR-V file from OpenCL using this new front-end compiler and after that I'm going to fix this bug.

Sam, are you OK with switching from SPIR 1.2 ang SPIR 2.0 generators to new clang?

yxsamliu:
I think we'd better find a way to support blocks in OpenCL.

[...]

PR #209, opened 2017-04-29
Adds handling for llvm.memmove-intrinsic

### LINKED ISSUE #205: Intrinsic llvm.memmove not supported
Trying to convert following code to SPIR-V throws an exception.

OpenCL-C code:

    struct SomeStruct
    {
	float16 f;
	int i;
    };

    __kernel void test_struct(const __global struct SomeStruct* in, __global struct SomeStruct* out)
    {
	struct SomeStruct tmp = *in;
	*out = tmp;
    }

OpenCL-C to LLVM-IR compiled with SPIRV-LLVM/SPIR (`clang version 3.6.1 (https://github.com/KhronosGroup/SPIR d7e44c3b27581e54ca0e522987d1ade2bd29b70d) (https://github.com/KhronosGroup/SPIRV-LLVM.git fffc52b7fb3552c045c398d68e22216c8d770c00)`)
LLVM-IR code (excerpt):

    %struct.SomeStruct = type { <16 x float>, i32, [60 x i8] }
    define spir_kernel void @test_struct(%struct.SomeStruct addrspace(1)* nocapture readonly %in, %struct.SomeStruct addrspace(1)* nocapture %out) #0 {
      %1 = bitcast %struct.SomeStruct addrspace(1)* %in to i8 addrspace(1)*
      %2 = bitcast %struct.SomeStruct addrspace(1)* %out to i8 addrspace(1)*
      call void @llvm.memmove.p1i8.p1i8.i32(i8 addrspace(1)* %2, i8 addrspace(1)* %1, i32 128, i32 64, i1 false)
      ret void
    }

    declare void @llvm.memmove.p1i8.p1i8.i32(i8 addrspace(1)* nocapture, i8 addrspace(1)* nocapture readonly, i32, i32, i1)

As you can see, this should be a simple mem-copy, but `llvm-spirv`throws:

    InvalidFunctionCall: Unexpected llvm intrinsic:  [Src: /opt/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:1314  ]
    Fails to save LLVM as SPIRV: InvalidFunctionCall: Unexpected llvm intrinsic:  [Src: /opt/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:1314  ]

doe300:
Closes #205.

Produces following SPIR-V output for the example given in #205:

    119734787 65536 393230 22 0 
    2 Capability Addresses 
    2 Capability Kernel 
    2 Capability Vector16 
    2 Capability Int8 
    5 ExtInstImport 1 "OpenCL.std"
    3 MemoryModel 1 2 
    6 EntryPoint 6 12 "test_struct"
    3 Source 3 102000 
    7 Name 3 "struct.SomeStruct"
    3 Name 13 "in"
    3 Name 14 "out"
    4 Decorate 21 FuncParamAttr 5 
    2 DecorationGroup 21 
    4 Decorate 13 FuncParamAttr 6 
    4 GroupDecorate 21 13 14 
    4 TypeInt 6 32 0 
    4 TypeInt 8 8 0 
    4 Constant 6 7 60 
    4 Constant 6 20 128 
    2 TypeVoid 2 
    3 TypeFloat 4 32 
    4 TypeVector 5 4 16 
    4 TypeArray 9 8 7 
    5 TypeStruct 3 5 6 9 
    4 TypePointer 10 5 3 
    5 TypeFunction 11 2 10 10 
    4 TypePointer 16 5 8 
    
    5 Function 2 12 0 11 
    3 FunctionParameter 10 13 
    3 FunctionParameter 10 14 
    
    2 Label 15 
    4 Variable 2 19 7 
    4 Bitcast 16 17 13 
    4 Bitcast 16 18 14 
    3 LifetimeStart 19 128 
    6 CopyMemorySized 19 17 20 2 64 
    6 CopyMemorySized 18 19 20 2 64 
    3 LifetimeStop 19 128 
    1 Return 
    
    1 FunctionEnd 

This result is "correct" in the sense, that the code does the right thing, but:
- [x] The temporary Variable is of type `void` and has no size specified.
- [x] We probably should add `LifetimeStart` and `LifeTimeStop` before/after the `CopyMemorySized` instruction.
- [x] According to the [SPIR-V specification, section 2.4](https://www.khronos.org/registry/spir-v/specs/1.1/SPIRV.html#_a_id_logicallayout_a_logical_layout_of_a_module):
> All OpVariable instructions in a function must be in the first block in the function. These instructions, together with any immediately preceding OpLine instructions, must be the first instructions in that block. (Note the validation rules prevent OpPhi instructions in the first block of a function.)

### DIFF
--- a/lib/SPIRV/SPIRVLowerMemmove.cpp
+++ b/lib/SPIRV/SPIRVLowerMemmove.cpp
@@ -1,0 +1,139 @@
+//===- SPIRVLowerMemmove.cpp - Lower llvm.memmove to llvm.memcpys ----------===//
+//
+//                     The LLVM/SPIRV Translator
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+// Copyright (c) 2014 Advanced Micro Devices, Inc. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal with the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimers.
+// Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimers in the documentation
+// and/or other materials provided with the distribution.
+// Neither the names of Advanced Micro Devices, Inc., nor the names of its
+// contributors may be used to endorse or promote products derived from this
+// Software without specific prior written permission.
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH
+// THE SOFTWARE.
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements lowering llvm.memmove into several llvm.memcpys.
+//
+//===----------------------------------------------------------------------===//
+#define DEBUG_TYPE "spvmemmove"
+
+#include "SPIRVInternal.h"
+#include "llvm/IR/InstVisitor.h"
+#include "llvm/IR/IntrinsicInst.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/Module.h"
+#include "llvm/IR/Verifier.h"
+#include "llvm/Pass.h"
+#include "llvm/PassSupport.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/raw_ostream.h"
+
+using namespace llvm;
+using namespace SPIRV;
+
+namespace SPIRV {
+cl::opt<bool> SPIRVLowerMemmoveValidate("spvmemmove-validate",
+    cl::desc("Validate module after lowering llvm.memmove instructions into " 
+        "llvm.memcpy"));
+
+class SPIRVLowerMemmove: public ModulePass,
+  public InstVisitor<SPIRVLowerMemmove> {
+public:
+  SPIRVLowerMemmove():ModulePass(ID), Context(nullptr) {
+    initializeSPIRVLowerMemmovePass(*PassRegistry::getPassRegistry());
+  }
+  virtual void visitMemMoveInst(MemMoveInst &I) {
+    IRBuilder<> Builder(I.getParent());
+    Builder.SetInsertPoint(&I);
+    auto *Dest = I.getRawDest();
+    auto *Src = I.getRawSource();
+    auto *SrcTy = Src->getType();
+    if (!isa<ConstantInt>(I.getLength())) 
+        // ToDo: for non-constant length, could use a loop to copy a 
+        // fixed length chunk at a time. For now simply fail
+        report_fatal_error("llvm.memmove of non-constant length not supported", 
+            false);
+    auto *Length = cast<ConstantInt>(I.getLength());
+    if (isa<BitCastInst>(Src)) 
+        // The source could be bit-cast from another type,
+        // need the original type for the allocation of the temporary variable
+        SrcTy = cast<BitCastInst>(Src)->getOperand(0)->getType();
+    auto Align = I.getAlignment();
+    auto Volatile = I.isVolatile();
+    Value *NumElements = nullptr;
+    uint64_t ElementsCount = 1;
+    if (SrcTy->isArrayTy()) {
+        NumElements = Builder.getInt32(SrcTy->getArrayNumElements());
+        ElementsCount = SrcTy->getArrayNumElements();
+    }
+    if (Mod->getDataLayout()->getTypeSizeInBits(SrcTy->getPointerElementType()) 
+        * ElementsCount !=  Length->getZExtValue() * 8)
+        report_fatal_error("Size of the memcpy should match the allocated memory", 
+            false);
+
+    auto *Alloca = Builder.CreateAlloca(SrcTy->getPointerElementType(), 
+        NumElements);
+    auto *LifetimeStart = Builder.CreateLifetimeStart(Alloca);
+    auto *FirstCpy = Builder.CreateMemCpy(Alloca, Src, Length, Align, Volatile);
+    auto *SecondCpy = Builder.CreateMemCpy(Dest, Alloca, Length, Align, 
+        Volatile);
+    auto *LifetimeEnd = Builder.CreateLifetimeEnd(Alloca);
+
+    SecondCpy->takeName(&I);
+    I.replaceAllUsesWith(SecondCpy);
+    I.dropAllReferences();
+    I.eraseFromParent();
+  }
+  virtual bool runOnModule(Module &M) {
+    Context = &M.getContext();
+    Mod = &M;
+    visit(M);
+
+    if (SPIRVLowerMemmoveValidate) {
+      DEBUG(dbgs() << "After SPIRVLowerMemmove:\n" << M);
+      std::string Err;
+      raw_string_ostream ErrorOS(Err);
+      if (verifyModule(M, &ErrorOS)){
+        Err = std::string("Fails to verify module: ") + Err;
+        report_fatal_error(Err.c_str(), false);
+      }
+    }
+    return true;
+  }
+
+  static char ID;
+private:
+  LLVMContext *Context;
+  Module *Mod;
+};
+
+char SPIRVLowerMemmove::ID = 0;
+}
+
+INITIALIZE_PASS(SPIRVLowerMemmove, "spvmemmove",
+    "Lower llvm.memmove into llvm.memcpy", false, false)
+
+ModulePass *llvm::createSPIRVLowerMemmove() {
+  return new SPIRVLowerMemmove();
+}
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1838,6 +1838,7 @@ addPassesForSPIRV(PassManager &PassMgr) {
   PassMgr.add(createSPIRVRegularizeLLVM());
   PassMgr.add(createSPIRVLowerConstExpr());
   PassMgr.add(createSPIRVLowerBool());
+  PassMgr.add(createSPIRVLowerMemmove());
 }
 
 bool
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -1878,8 +1878,7 @@ protected:
     assert(Obj->getStorageClass() == StorageClassFunction &&
         "Invalid storage class");
     assert(Obj->getType()->isTypePointer() &&
-        Obj->getType()->getPointerElementType()->isTypeInt() &&
-        "Object type must be an integer type scalar");
+        "Objects type must be a pointer");
     if (!Obj->getType()->getPointerElementType()->isTypeVoid() ||
         !Module->hasCapability(CapabilityAddresses))
       assert(Size == 0 && "Size must be 0");
--- a/include/llvm/Support/SPIRV.h
+++ b/include/llvm/Support/SPIRV.h
@@ -56,6 +56,7 @@ void initializeOCLTypeToSPIRVPass(PassRegistry&);
 void initializeSPIRVLowerBoolPass(PassRegistry&);
 void initializeSPIRVLowerConstExprPass(PassRegistry&);
 void initializeSPIRVLowerOCLBlocksPass(PassRegistry&);
+void initializeSPIRVLowerMemmovePass(PassRegistry&);
 void initializeSPIRVRegularizeLLVMPass(PassRegistry&);
 void initializeSPIRVToOCL20Pass(PassRegistry&);
 void initializeTransOCLMDPass(PassRegistry&);
@@ -134,6 +135,9 @@ ModulePass *createSPIRVLowerConstExpr();
 
 /// Create a pass for lowering OCL 2.0 blocks to functions calls.
 ModulePass *createSPIRVLowerOCLBlocks();
+
+/// Create a pass for lowering llvm.memmove to llvm.memcpys with a temporary variable.
+ModulePass *createSPIRVLowerMemmove();
 
 /// Create a pass for regularize LLVM module to be translated to SPIR-V.
 ModulePass *createSPIRVRegularizeLLVM();
--- a/lib/SPIRV/CMakeLists.txt
+++ b/lib/SPIRV/CMakeLists.txt
@@ -30,6 +30,7 @@ add_llvm_library(LLVMSPIRVLib
   SPIRVLowerBool.cpp
   SPIRVLowerConstExpr.cpp
   SPIRVLowerOCLBlocks.cpp
+  SPIRVLowerMemmove.cpp
   SPIRVReader.cpp
   SPIRVRegularizeLLVM.cpp
   SPIRVToOCL20.cpp
--- a/test/SPIRV/transcoding/llvm.memmove.ll
+++ b/test/SPIRV/transcoding/llvm.memmove.ll
@@ -1,0 +1,70 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.txt
+; RUN: FileCheck < %t.txt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV-NOT: llvm.memmove
+
+; CHECK-SPIRV: Variable {{[0-9]+}} [[mem:[0-9]+]] 7
+
+; CHECK-SPIRV: LifetimeStart [[mem]] [[size:[0-9]+]]
+; CHECK-SPIRV: Bitcast [[i8Ty:[0-9]+]] [[tmp1:[0-9]+]] [[mem]]
+; CHECK-SPIRV: CopyMemorySized [[tmp1]] {{[0-9]+}} {{[0-9]+}}
+; CHECK-SPIRV: Bitcast [[i8Ty]] [[tmp2:[0-9]+]] [[mem]]
+; CHECK-SPIRV: CopyMemorySized {{[0-9]+}} [[tmp2]] {{[0-9]+}}
+; CHECK-SPIRV: LifetimeStop [[mem]] [[size]]
+
+; CHECK-LLVM-NOT: llvm.memmove
+
+; CHECK-LLVM: [[local:%[0-9]+]] = alloca %struct.SomeStruct
+; CHECK-LLVM: [[tmp1:%[0-9]+]] = bitcast %struct.SomeStruct* [[local]] to [[type:i[0-9]+\*]]
+; CHECK-LLVM: call void @llvm.lifetime.start({{i[0-9]+}} {{-?[0-9]+}}, [[type]] [[tmp1]])
+; CHECK-LLVM: [[tmp2:%[0-9]+]] = bitcast %struct.SomeStruct* [[local]] to [[type]]
+; CHECK-LLVM: call void @llvm.memcpy
+; CHECK-LLVM:  ([[type]] [[tmp2]],
+; CHECK-LLVM:  {{i[0-9]+}} [[size:[0-9]+]]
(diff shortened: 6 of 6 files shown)


yxsamliu:
You can add SPIRVFunction::addVariable to make sure the variable instruction is properly ordered, then modify SPIRVModuleImpl::addVariable to call it.

doe300:
So, this version does everything correct, as far as I can tell and is ready to be pulled.

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -67,6 +67,12 @@ SPIRVFunctionParameter::foreachAttr(
   }
 }
 
+SPIRVVariable *SPIRVFunction::addVariable(SPIRVVariable *Var)
+{
+  //according to the SPIR-V specs, the Variable (with Storage Class Function) need to be at the beginning of the first block of the function, see section 2.4 (Logical Layout of a Module)
+  return BBVec.at(0)->addVariable(Var);
```
yxsamliu: needs to check if BBVec is empty. If so add a BB.

lib/SPIRV/libSPIRV/SPIRVBasicBlock.cpp
```diff
@@ -69,6 +69,15 @@ SPIRVBasicBlock::addInstruction(SPIRVInstruction *I) {
   return I;
 }
 
+SPIRVVariable *
+SPIRVBasicBlock::addVariable(SPIRVVariable *Var) {
+  assert(Var && "Invalid variable");
+  Module->add(Var);
+  Var->setParent(this);
+  InstVec.insert(InstVec.begin(), Var);
```
yxsamliu: This will cause auto vars in source code emitted in reverse order and break existing lit tests.

Should find the first non-Variable instruction and insert before that.

yxsamliu:
Please check the Travis CI build status. It should pass.

doe300:
Made the requested changes. Also now the tests compile.

I'm not quite sure though about adding the new BasicBlock in SPIRVFunction::addVariable (line 70). Do I have to specify an ID? If so, where do I get it from?

lib/SPIRV/libSPIRV/SPIRVBasicBlock.cpp
```diff
@@ -69,6 +69,23 @@ SPIRVBasicBlock::addInstruction(SPIRVInstruction *I) {
...
 }
 
+SPIRVVariable *
+SPIRVBasicBlock::addVariable(SPIRVVariable *Var) {
+  assert(Var && "Invalid variable");
+  Module->add(Var);
+  Var->setParent(this);
+  auto It = InstVec.begin();
+  while(It != InstVec.end())
+  {
```
bader: Minor style comment.
Please, follow the existing style - move { to the line above.
I think the whole loop can be replaced with standard find algorithm.

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -67,6 +67,16 @@ SPIRVFunctionParameter::foreachAttr(
   }
 }
 
+SPIRVVariable *SPIRVFunction::addVariable(SPIRVVariable *Var)
+{
```
bader: Please, move the brace to the line above.

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -67,6 +67,16 @@ SPIRVFunctionParameter::foreachAttr(
   }
 }
 
+SPIRVVariable *SPIRVFunction::addVariable(SPIRVVariable *Var)
+{
+  //according to the SPIR-V specs, the Variable (with Storage Class Function) need to be at the beginning of the first block of the function, see section 2.4 (Logical Layout of a Module)
+  if(getNumBasicBlock() == 0)
+  {
```
bader: Please, move the brace to the line above.

bader (commented):
It would be great if you add a regression test validating correct behavior. You can use the test case from the issue #205.
Just, a side note: backward translation will not be able to recover original llvm intrinsic.

doe300:
Applied the style-changes and added a test case. 

Is this test-case any good? If not, how should I write it?

test/SPIRV/llvm.memmove.ll
```diff
@@ -0,0 +1,50 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
+
+; CHECK-NOT: llvm.memmove
+
+; CHECK: Variable [[voidPtr:[0-9]+]] [[mem:[0-9]+]] 7
```
bader: There is no need in voidPtr variable - it's not used in the test. Suggest replacing with {{[0-9]+}}.

doe300:
The tests in DEBUG mode on travis CI fail with "Object type must be an integer type scalar" in SPIRVInstruction.h:1882.

I don't understand, why the pointer-type needs to be an integer. Especially, because the next line seems to accept the `void*` type.

bader:
It looks like a bug in the assertion statement.

Here is the quote from the [spec](https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#OpLifetimeStart):
"_Pointer_ is a pointer to the object whose lifetime is starting. Its type must be an OpTypePointer with Storage Class Function.

_Size_ must be 0 if Pointer is a pointer to a non-void type or the Addresses capability is not being used. If Size is non-zero, it is the number of bytes of memory whose lifetime is starting. Its type must be an integer type scalar. It is treated as unsigned; if its type has Signedness of 1, its sign bit cannot be set."

"Size" parameter type must be integer. "Pointer" can point to non-integer type values.

Could you fix the assertion statement, please?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -1877,7 +1877,6 @@ class SPIRVLifetime : public SPIRVInstruction {
     assert(Obj->getStorageClass() == StorageClassFunction &&
         "Invalid storage class");
     assert(Obj->getType()->isTypePointer() &&
-        Obj->getType()->getPointerElementType()->isTypeInt() &&
         "Object type must be an integer type scalar");
```
bader: Please, update the message to avoid confusion.
"Objects type must be a pointer".

bader (approved):
One minor comment.

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -67,6 +67,14 @@ SPIRVFunctionParameter::foreachAttr(
   }
 }
 
+SPIRVVariable *SPIRVFunction::addVariable(SPIRVVariable *Var) {
+  //according to the SPIR-V specs, the Variable (with Storage Class Function) need to be at the beginning of the first block of the function, see section 2.4 (Logical Layout of a Module)
+  if(getNumBasicBlock() == 0) {
```
yxsamliu: insert space after `if`

yxsamliu:
There is still travis-ci build failure. Looking at the result:

FAIL: LLVM :: SPIRV/llvm.memmove.ll (8805 of 12438)
******************** TEST 'LLVM :: SPIRV/llvm.memmove.ll' FAILED ********************
Script:
--
/home/travis/build/KhronosGroup/SPIRV-LLVM/build/./bin/llvm-as /home/travis/build/KhronosGroup/SPIRV-LLVM/test/SPIRV/llvm.memmove.ll -o /home/travis/build/KhronosGroup/SPIRV-LLVM/build/test/SPIRV/Output/llvm.memmove.ll.tmp.bc
/home/travis/build/KhronosGroup/SPIRV-LLVM/build/./bin/llvm-spirv /home/travis/build/KhronosGroup/SPIRV-LLVM/build/test/SPIRV/Output/llvm.memmove.ll.tmp.bc -spirv-text -o - | /home/travis/build/KhronosGroup/SPIRV-LLVM/build/./bin/FileCheck /home/travis/build/KhronosGroup/SPIRV-LLVM/test/SPIRV/llvm.memmove.ll
--
Exit Code: 2

Command Output (stderr):
--
llvm-spirv: /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/libSPIRV/SPIRVInstruction.h:1880: virtual void SPIRV::SPIRVLifetime<256>::validate() const [OC = 256]: Assertion `Obj->getType()->isTypePointer() && "Objects type must be a pointer"' failed.
Stack dump:
0.	Program arguments: /home/travis/build/KhronosGroup/SPIRV-LLVM/build/./bin/llvm-spirv /home/travis/build/KhronosGroup/SPIRV-LLVM/build/test/SPIRV/Output/llvm.memmove.ll.tmp.bc -spirv-text -o - 
1.	Running pass 'LLVMToSPIRV' on module '/home/travis/build/KhronosGroup/SPIRV-LLVM/build/test/SPIRV/Output/llvm.memmove.ll.tmp.bc'.
#0 0x80b21e llvm::sys::PrintStackTrace(_IO_FILE*) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/Support/Unix/Signals.inc:422:15
#1 0x80bfdb PrintStackTraceSignalHandler(void*) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/Support/Unix/Signals.inc:481:1
#2 0x80e704 SignalHandler(int) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/Support/Unix/Signals.inc:198:60
#3 0x2adbb96de330 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x10330)
#4 0x2adbba35bc37 gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x36c37)
#5 0x2adbba35f028 abort (/lib/x86_64-linux-gnu/libc.so.6+0x3a028)
#6 0x2adbba354bf6 (/lib/x86_64-linux-gnu/libc.so.6+0x2fbf6)
#7 0x2adbba354ca2 (/lib/x86_64-linux-gnu/libc.so.6+0x2fca2)
#8 0x4a85f6 SPIRV::SPIRVLifetime<(spv::Op)256>::validate() const /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/libSPIRV/SPIRVInstruction.h:1881:10
#9 0x498c4d SPIRV::SPIRVLifetime<(spv::Op)256>::SPIRVLifetime(unsigned int, unsigned int, SPIRV::SPIRVBasicBlock*) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/libSPIRV/SPIRVInstruction.h:1861:5
#10 0x498b6d SPIRV::SPIRVModuleImpl::addLifetimeInst(spv::Op, SPIRV::SPIRVValue*, unsigned int, SPIRV::SPIRVBasicBlock*) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/libSPIRV/SPIRVModule.cpp:1125:12
#11 0x5919d9 SPIRV::LLVMToSPIRV::transIntrinsicInst(llvm::IntrinsicInst*, SPIRV::SPIRVBasicBlock*) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:1313:5
#12 0x58f1b7 SPIRV::LLVMToSPIRV::transValueWithoutDecoration(llvm::Value*, SPIRV::SPIRVBasicBlock*, bool) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:1157:22
#13 0x58a1b3 SPIRV::LLVMToSPIRV::transValue(llvm::Value*, SPIRV::SPIRVBasicBlock*, bool) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:812:13
#14 0x5938b2 SPIRV::LLVMToSPIRV::transFunction(llvm::Function*) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:1486:7
#15 0x593d8b SPIRV::LLVMToSPIRV::translate() /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:1530:5
#16 0x5a047f SPIRV::LLVMToSPIRV::runOnModule(llvm::Module&) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:178:5
#17 0x73da7c (anonymous namespace)::MPPassManager::runOnModule(llvm::Module&) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/IR/LegacyPassManager.cpp:1616:23
#18 0x73d65e llvm::legacy::PassManagerImpl::run(llvm::Module&) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/IR/LegacyPassManager.cpp:1723:16
#19 0x73e041 llvm::legacy::PassManager::run(llvm::Module&) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/IR/LegacyPassManager.cpp:1756:10
#20 0x5967d1 llvm::WriteSPIRV(llvm::Module*, llvm::raw_ostream&, std::string&) /home/travis/build/KhronosGroup/SPIRV-LLVM/lib/SPIRV/SPIRVWriter.cpp:1857:3
#21 0x409828 convertLLVMToSPIRV() /home/travis/build/KhronosGroup/SPIRV-LLVM/tools/llvm-spirv/llvm-spirv.cpp:154:8
#22 0x408fd8 main /home/travis/build/KhronosGroup/SPIRV-LLVM/tools/llvm-spirv/llvm-spirv.cpp:312:12
#23 0x2adbba346f45 __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21f45)
#24 0x4088e4 _start (/home/travis/build/KhronosGroup/SPIRV-LLVM/build/bin/llvm-spirv+0x4088e4)

Can  you fix that?

doe300:
I can't figure out, why this test fails. So, besides removing the test I don't know of any fix.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1303,6 +1303,28 @@ LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II, SPIRVBasicBlock *BB) {
       transValue(II->getOperand(2), BB),
       getMemoryAccess(cast<MemIntrinsic>(II)),
       BB);
+  case Intrinsic::memmove: {
+    //Since memory areas for source and dest can overlap, 1. copy to temporary, 2. copy to destination
+    SPIRVType *Ty = transType(II->getType());
```
yxsamliu: should be II->getOperand(0)->getType()

test/SPIRV/llvm.memmove.ll
```diff
@@ -0,0 +1,50 @@
+; RUN: llvm-as %s -o %t.bc
```
yxsamliu: can we make it a transcodeing test to make sure the generated SPIRV can be translated back to LLVM IR?

yxsamliu:
I think I know why. The type passed to addVariable is wrong. I've added inline comments about that.

yxsamliu:
Also, can you add an assertion to SPIRVVariable::validate to assert the type is pointer type? Thanks.

doe300:
Applied the requested changes, the trancoding-test fails, since the method-signatures do not match for the `llvm.memcpy` calls:

    Fails to verify module: Call parameter type does not match function signature!
      %1 = alloca i8
     i8 addrspace(1)*  call void @llvm.memcpy.p1i8.p1i8.i32(i8* %1, i8 addrspace(1)* %2, i32 128, i32 64, i1 false)
    Call parameter type does not match function signature!
      %1 = alloca i8
     i8 addrspace(1)*  call void @llvm.memcpy.p1i8.p1i8.i32(i8 addrspace(1)* %3, i8* %1, i32 128, i32 64, i1 false)

With the single function declaration for `llvm.memcpy`:

    declare void @llvm.memcpy.p1i8.p1i8.i32(i8 addrspace(1)* nocapture, i8 addrspace(1)* nocapture readonly, i32, i32, i1) #0

As far as i can tell, there should be 2 declarations, something like:

    declare void @llvm.memcpy.p1i8.p1i8.i32(i8*, i8 addrspace(1)* nocapture readonly, i32, i32, i1)
    declare void @llvm.memcpy.p1i8.p1i8.i32(i8 addrspace(1)* nocapture, i8* readonly, i32, i32, i1)

yxsamliu:
It seems to be more complicated than expected. One issue is the type of the temporary variable. It should be either a pointer to the object type to be moved, or a pointer to an array of chars which have the same size of the memmov operation.

Another issue is that adding an existing type to SPIR-V module may cause problem. The SPIR-V module does not have a fold set to guarantee uniqueness of types, so the user has to make sure no duplicate types are added to SPIR-V module.

For this type of transformation, it may be easier to write a pass like SPIRVLowerBool to transform memmov to memcpy in LLVM IR.

AlexeySotkin:
Using -O3 we transform two memcpy calls to single memmove call. For translation to SPIR-V purposes it looks reasonable to transform the memmove back to memcpy. 

Maybe we just should not use -O3 in the first place?
Moreover I don't see any benefits of using -O3, if we end up with the same SPIR-V code as we had with -O0.

yxsamliu:
That can be a workaround if users are using Clang. I think it is better to be able to lower memmov.

doe300:
I need the other optimizations run with `-O3`, so disabling optimizations is not an option.
I rewrote the pull-request using a lowering pass as suggested by @yxsamliu .

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -0,0 +1,118 @@
...
+    initializeSPIRVLowerMemmovePass(*PassRegistry::getPassRegistry());
+  }
+  virtual void visitMemMoveInst(MemMoveInst &I) {
+    IRBuilder<> Builder(I.getParent());
+    Builder.SetInsertPoint(&I);
+    auto Dest = I.getRawDest();
+    auto Src = I.getRawSource();
+    auto SrcTy = Src->getType();
+    ConstantInt *Length = nullptr;
+    if(isa<ConstantInt>(I.getLength())) //the length (in bytes) could be non-constant
```
yxsamliu: space between if and (

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -0,0 +1,118 @@
...
+class SPIRVLowerMemmove: public ModulePass,
+  public InstVisitor<SPIRVLowerMemmove> {
+public:
+  SPIRVLowerMemmove():ModulePass(ID), Context(nullptr) {
+    initializeSPIRVLowerMemmovePass(*PassRegistry::getPassRegistry());
+  }
+  virtual void visitMemMoveInst(MemMoveInst &I) {
+    IRBuilder<> Builder(I.getParent());
+    Builder.SetInsertPoint(&I);
+    auto Dest = I.getRawDest();
```
yxsamliu: For pointers better use auto *. Same as below.

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -0,0 +1,118 @@
...
+    auto Src = I.getRawSource();
+    auto SrcTy = Src->getType();
+    ConstantInt *Length = nullptr;
+    if(isa<ConstantInt>(I.getLength())) //the length (in bytes) could be non-constant
+        Length = ConstantInt::get(Type::getInt64Ty(*Context), cast<ConstantInt>(I.getLength())->getZExtValue());
```
yxsamliu: Can you just use I.getLength() instead of create a new ConstantInt?

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -0,0 +1,118 @@
...
+    if(isa<BitCastInst>(Src)) //the source could be bit-cast from another type, need the original type for the allocation
+        SrcTy = cast<BitCastInst>(Src)->getOperand(0)->getType();
+    auto Align = I.getAlignment();
+    auto Volatile = I.isVolatile();
+
+    auto Alloca = Builder.CreateAlloca(SrcTy->getPointerElementType(), SrcTy->isArrayTy() ? Builder.getInt32(SrcTy->getArrayNumElements()) : nullptr);
```
yxsamliu: this line may exceed 80 columns.
Please follow http://llvm.org/docs/CodingStandards.html

lib/SPIRV/SPIRVLowerMemmove.cpp
yxsamliu: If it is not constant we can create a loop and each time copy one chunk of fixed length.

However, for now, maybe just report_fatal_error as "llvm.memmov of non-constant length not supported".

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -0,0 +1,118 @@
...
+    Builder.SetInsertPoint(&I);
+    auto Dest = I.getRawDest();
+    auto Src = I.getRawSource();
+    auto SrcTy = Src->getType();
+    ConstantInt *Length = nullptr;
+    if(isa<ConstantInt>(I.getLength())) //the length (in bytes) could be non-constant
+        Length = ConstantInt::get(Type::getInt64Ty(*Context), cast<ConstantInt>(I.getLength())->getZExtValue());
+    if(isa<BitCastInst>(Src)) //the source could be bit-cast from another type, need the original type for the allocation
+        SrcTy = cast<BitCastInst>(Src)->getOperand(0)->getType();
+    auto Align = I.getAlignment();
```
yxsamliu: need to check the size of the variable matches the length of memmov. If not report_fatal_error.

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -105,6 +125,7 @@ class SPIRVLowerMemmove: public ModulePass,
   static char ID;
 private:
   LLVMContext *Context;
+  Module *Mod;
```
yxsamliu: Is this member used?

doe300: line 90 to query the size of the allocated type: `Mod->getDataLayout()->getTypeSizeInBits(SrcTy->getPointerElementType())`. I created the member, since I didn't know how else to get the Module instance.

yxsamliu: Sorry I missed that. Then it looks fine.

yxsamliu:
There are too many commits in this pull request. Can you squash them into one commit? Otherwise LGTM.

PR #214, opened 2017-05-10
Khronos/spirv 3.8

abel-bernabeu:


abel-bernabeu:
Please pull two fixes: one for an issue I introduced myself and another for a unit test that was not passing on the 3.8.1 branch.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1,1448 +1,1448 @@
-//===- SPIRVUtil.cpp - SPIR-V Utilities -------------------------*- C++ -*-===//
```
yxsamliu: It seems everything is treated as changed. Is it due to line ending change?

abel-bernabeu:
Yes, it is the line endings. I only intend to change one line. Let me fix
it before you pull

PR #217, opened 2017-05-16
Changing translation of llvm.memset intrinsic

### LINKED ISSUE #203: SPIRV Writer fails with "Can't translate llvm.memset with non-zero value argument"
I see this assertion when I use `-emit-spirv` with these files:
https://github.com/KhronosGroup/SPIR/blob/spirv-1.1/test/OpenCL/OpenCL22/attributes/TestBasicAttributes_kernel.cl
https://github.com/KhronosGroup/libclcxx/blob/master/test/libclcxx/address_spaces/TestAddrSpaceTypeConstantPtr.cl
https://github.com/KhronosGroup/libclcxx/blob/master/test/libclcxx/address_spaces/TestAddrSpaceTypeGlobalPtr.cl
https://github.com/KhronosGroup/libclcxx/blob/master/test/libclcxx/address_spaces/TestAddrSpaceTypeLocalPtr.cl
https://github.com/KhronosGroup/libclcxx/blob/master/test/libclcxx/address_spaces/TestAddrSpaceTypePrivatePtr.cl

```
clang -cc1 -emit-spirv -triple=spir-unknown-unknown -cl-std=c++ assert-memset.cl
clang: .../llvm/lib/SPIRV/SPIRVWriter.cpp:1296: SPIRV::SPIRVValue* SPIRV::LLVMToSPIRV::transIntrinsicInst(llvm::IntrinsicInst*, SPIRV::SPIRVBasicBlock*):
Assertion `!"Can't translate llvm.memset with non-zero value argument"' failed.
```

Here my test case:
```cpp
struct S1
{
    int x;
    int y;
};

S1 foo11()
{
    return S1();
}

S1 foo12()
{
    return S1{};
}

S1 foo13()
{
    S1 s;
    return s;
}

class S2
{
    int x;
    int y;
};

S2 foo21()
{
    return S2();
}

S2 foo22()
{
    return S2{};
}

S2 foo23()
{
    S2 s;
    return s;
}
```

foo11 - fail
foo12 - ok
foo13 - ok

foo21 - fail
foo22 - fail
foo23 - ok

Full stack trace:
[assert-memset.txt](https://github.com/KhronosGroup/SPIRV-LLVM/files/725819/assert-memset.txt)
LLVM:
[assert-memset.ll.txt](https://github.com/KhronosGroup/SPIRV-LLVM/files/725828/assert-memset.ll.txt)

(GCC 6.3.1 on linux and VS2015 on windows in Debug)

AlexeySotkin:
In most cases llvm.memset is emitted by Clang to do zero-initialization
in default constructors. With this commit this case is handled as followed:
1. Create global variable initialized by a null constant array. Type of
   array's element is i8, length of the array is `len` operand of the
   llvm.memset intrinsic.
2. Copy the array to the target by OpCopyMemorySized, keeping alignment and
   isvolatile arguments.

Translation from SPIR-V to LLVM is updated as well.
Fixes #203


AlexeySotkin:
I think addition of memset and memmove instructions to the SPIR-V specification would be a better solution.

PR #222, opened 2017-10-10
~SPIRVModuleImpl() crashes when delete no-id entries

lifpan:
It crashes because it tries to delete some no-id entry (OpStore) twice, which is
pushed into vector "EntryNoId" twice.

Do not add entries of SPIRVBasicBlock and SPIRVInstruction into Module when call
Decoder.getEntry(). They will be added at addBasicBlock() and addInstruction().

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -378,15 +378,12 @@ private:
 };
 
 SPIRVModuleImpl::~SPIRVModuleImpl() {
-  //ToDo: Fix bug causing crash
-  //for (auto I:IdEntryMap)
-  //  delete I.second;
+  for (auto I:IdEntryMap)
+    delete I.second;
 
-  // ToDo: Fix bug causing crash
-  //for (auto I:EntryNoId) {
-  //  bildbgs() << "[delete] " << *I;
-  //  delete I;
-  //}
+  for (auto I:EntryNoId) {
+    delete I;
+  }
 
   for (auto C : CapMap)
     delete C.second;
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -332,7 +332,7 @@ private:
   SPIRVMemoryModelKind MemoryModel;
 
   typedef std::map<SPIRVId, SPIRVEntry *> SPIRVIdToEntryMap;
-  typedef std::vector<SPIRVEntry *> SPIRVEntryVector;
+  typedef std::set<SPIRVEntry *> SPIRVEntrySet;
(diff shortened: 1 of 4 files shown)


bader:
Thanks for the patch, but it looks like it doesn't resolve the issue in all cases.
Please, take a look at the Trevis results. There are failing tests, which crash in ~SPIRVModuleImpl().

lifpan:
Where to get the test and how to run it at my local?

bader:
@lifpan, tests are located in test/SPIRV/ directory. 
There are several ways to run these tests, but easiest way to validate your patch is to build 'check-all' target e.g. 'make check-all'.

yxsamliu:
you can also use llvm-lit -v followed by the SPIR-V test directory, which should be at llvm/test/SPIRV.

lifpan:
Fix code for OpGroupDecorate, OpLine, and OpUndef.
Pass llvm-lit -v test/SPIRV:
  Expected Passes    : 111
  Expected Failures  : 1

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -552,7 +553,8 @@ SPIRVModuleImpl::addEntry(SPIRVEntry *Entry) {
     } else
       IdEntryMap[Id] = Entry;
   } else {
-    EntryNoId.push_back(Entry);
+    if (EntryNoId.empty() || EntryNoId.back() != Entry)
```
AlexeySotkin: Why this condition? Probably we can use std::set to avoid duplicates.

lifpan: For example, entry of OpGroupDecorate will be added twice when decode SPIRV binary:
1) SPIRVGroupDecorateGeneric::decode(std::istream &I) -> Module->addGroupDecorateGeneric(this)
2) Decoder.getEntry() -> M.add(Entry)

AlexeySotkin: As far as  I understand with this condition the entry will be added to the vector if 1) the vector is empty or 2) if the last element of the vector equals to the entry. What if the entry is already in the vector but it is not at the end of the vector?

yxsamliu:
LGTM. Thanks.

lifpan:
@yxsamliu I found a mistake near line 1508 in SPIRVModule.cpp, that Decoder.getEntry() may return NULL. This can happen if the binary is not strictly encoded.

Please help correct it:
   while(Decoder.getWordCountAndOpCode())
     M.add(Decoder.getEntry());
-->
  while (Decoder.getWordCountAndOpCode()) {
    SPIRVEntry *Entry = Decoder.getEntry();
    if (Entry != NULL)
      M.add(Entry);
  }

bader:
@lifpan, could you create another pull request and attach the test case, which reproduce the issue, please?

PR #223, opened 2017-10-18
~SPIRVModuleImpl() crashes when delete id entries

lifpan:
A forwarded entry has 2 duplicate elements in "IdEntryMap", one element is
at map slot of old id, and another is at forward id.


bader:
LGTM. Thanks!

AlexeySotkin:
LGTM, Thanks!

yxsamliu:
Good catch. Thanks!

PR #224, opened 2017-10-20
Decoder.getEntry() may returns NULL

lifpan:
This happens when SPIR-V binary contains useless word which does not belong to
any instruction. Its high 16-bit is 0, and the decoder discards it.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -1506,7 +1506,7 @@ operator>> (std::istream &I, SPIRVModule &M) {
 
   while (Decoder.getWordCountAndOpCode()) {
     SPIRVEntry *Entry = Decoder.getEntry();
-    if (Entry != NULL)
+    if (Entry != nullptr)
       M.add(Entry);
   }
 
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -1504,8 +1504,11 @@ operator>> (std::istream &I, SPIRVModule &M) {
   Decoder >> MI.InstSchema;
   assert(MI.InstSchema == SPIRVISCH_Default && "Unsupported instruction schema");
 
-  while(Decoder.getWordCountAndOpCode())
-    M.add(Decoder.getEntry());
+  while (Decoder.getWordCountAndOpCode()) {
+    SPIRVEntry *Entry = Decoder.getEntry();
+    if (Entry != NULL)
+      M.add(Entry);
+  }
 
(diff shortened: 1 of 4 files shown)


lifpan:
Here is one SPIR-V binary generated by glslangValidator.exe (KhronosGroup/glslang), and the original GLSL shader.
[ObjInput_TestIndirect.zip](https://github.com/KhronosGroup/SPIRV-LLVM/files/1400903/ObjInput_TestIndirect.zip)

llvm-spirv -r ObjInput_TestIndirect.spv

llvm-spirv does not accept "GLSL.std.450", so please manually skip it:
          %1 = OpExtInstImport "GLSL.std.450"

bader:
@lifpan, is it possible to add the test to SPIR-V lit test suite?
Here is an example of the SPIR-V test in this test suite:
https://github.com/KhronosGroup/SPIRV-LLVM/blob/khronos/spirv-3.6.1/test/SPIRV/vector_times_scalar.spt

Note: you should put run commands at the end of SPIR-V text file.
Probably you don't need to run FileCheck - only llvm-spirv.

test/SPIRV/useless_word.spt
```diff
@@ -0,0 +1,35 @@
+119734787 65536 393230 11 0 
+2 Capability Addresses 
+0 Capability
```
bader: Just out of curiosity. Which instruction causes the issue?
```
0 Capability
```
or 
```
0 Nop
```
Both?

bader (approved):
LGTM.

test/SPIRV/useless_word.spt
lifpan: Both. By the way, shall we name it "redundant_word.spt"?

bader: Sound good to me.
Feel free to update the request.
I'll let Alexey S and Sam to take a look before the merge.
Thanks!

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1504,8 +1504,11 @@ operator>> (std::istream &I, SPIRVModule &M) {
   Decoder >> MI.InstSchema;
   assert(MI.InstSchema == SPIRVISCH_Default && "Unsupported instruction schema");
 
-  while(Decoder.getWordCountAndOpCode())
-    M.add(Decoder.getEntry());
+  while (Decoder.getWordCountAndOpCode()) {
+    SPIRVEntry *Entry = Decoder.getEntry();
+    if (Entry != NULL)
```
AlexeySotkin: It is common to use nullptr instead of NULL here. Please update.
Other than that, LGTM. Thanks.

yxsamliu:
LGTM. Thanks!

PR #227, opened 2018-02-06
[SPIRV] Fix capabilities for Atomic instructions

AlexeySotkin:


lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2044,10 +2044,32 @@ _SPIRV_OP(GroupCommitReadPipe, false, 6)
...
 
-class SPIRVAtomicInstBase:public SPIRVInstTemplateBase {
+class SPIRVAtomicInstBase : public SPIRVInstTemplateBase {
 public:
   SPIRVCapVec getRequiredCapability() const override {
-    return getVec(CapabilityInt64Atomics);
+    SPIRVCapVec CapVec;
+    // Most of atomic instructions do not require any capabilities
+    // ... unless they operate on 64-bit integers.
+    if (hasType() && getType()->isTypeInt(64)) {
```
bader, line 2053: Probably just to simplify the implementation we can process both cases in setOpWords method, where we have all information we need to decide whether we need Int64Atomics capability or not.

bader (approved):
Other than one very minor comment looks good to me.
Thanks for very informative comments.

Issue #228, opened 2018-02-20
OpenCL file compiled by Khronos OpenCL Clang compiler+SPIRV-LLVM using cl_intel_subgroups+cl_intel_subgroups_short omits needed OpExtension "SPV_INTEL_subgroups" and also Vector16 capability in generated SPIR-V


oscarbg:
Hi,
@bashbaug @AlexeySotkin @yxsamliu
the issue in fact is detailed here:
https://github.com/KhronosGroup/SPIRV-Tools/issues/1327
can you please take a detailed look.. you have the OpenCL cl file, the compiler arguments and the resulting spirv file and output from spirv-val..

bashbaug:
Thanks for the report.  I'll take a look.

oscarbg:
let me ask some somewhat related questions:
don't tested yet but it's that SPIR-V shaders usable on any public Intel OpenCL driver either on Windows or Linux platform..
also how is related to new Intel OpenCL Neo driver (https://github.com/intel/compute-runtime)..
it supports SPV intel subgroups already?
just opened an issue just aksing for info here:
https://github.com/intel/compute-runtime/issues/13
perhaps can contribute there..
and finally one last question:
I'm having trouble compiling the OpenCL file attached below (Opt_MatrixMultiplication.cl) to SPIR-V of this sample obtained from Intel site using the same Khronos OpenCL Clang compiler+SPIRV-LLVM I compiled from master that has "SPV_INTEL_subgroups":
[GEMM_sample_Jun2017.zip](https://github.com/KhronosGroup/SPIRV-LLVM/files/1742242/GEMM_sample_Jun2017.zip)

it uses cl_intel_subgroups but I see in code also something like:
#ifdef cl_intel_simd_operations_placeholder
#define cl_intel_subgroups
#define sub_group_broadcast             intel_simd_shuffle
#endif

so is this OpenCL file compilable yet to SPIR-V..
I say because it seems the fastest GEMM code avaiable for Intel HD Graphics right now and also with some very optimized f16 matmal code..
would be happy if you can take a look and comment..

bashbaug:
I had a look and here's what I found.

[...]

PR #229, opened 2018-02-28
add the Vector16 capability for both vec8 and vec16 types

bashbaug:
This is a partial fix for #228.  Even though the name of the capability is Vector16, it needs to be added for both vectors or 8 elements and vectors of 16 elements.

CC: @bader


bader (approved):
It would be nice to have a regression test.

bashbaug:
Good idea - I added a regression test for this case.

PR #230, opened 2018-03-06
Formatting spirv-6.0 to follow LLVM coding style guidelines

pierremoreau:
Changes made:

* Make sure all files use UNIX-style line endings instead of a mix of UNIX-style and Windows-style;
* Run clang-format and clang-tidy to ensure the codebase follows LLVM coding style rather than a mix of different coding styles;
* Run clang-tidy to address some of the warnings, such as missing 'override' keywords, removing explicit `std::move()` preventing copy elision optimisations;
* Add a missing include to SPIRVEnum.h in SPIRVExtInst.h, fixing a compilation error with clang 5.0.

Those results in many lines being changed, but given that there is no history: 1) it is not disrupting the history, and 2) those patches could be squashed into https://github.com/KhronosGroup/SPIRV-LLVM/commit/f00a6386a210483c85eaeb8bd68920ab236961ee without any issues, giving a clean state that conforms to LLVM coding style guidelines.

Note: Some of the headers would need some updates, which I haven’t done:
1. the first line of some headers is longer than the expected text width, and
2. that same first line sometimes uses a ‘-’ character which is displayed as “<96>” by Git/Vim (wrong encoding maybe?).


yxsamliu:
the tests failed due to timeout. @AlexeySotkin is it possible to change the timeout value?

AlexeySotkin:
LGTM, thanks, but I'm not sure that this is a branch we will proceed with.

PR #233, opened 2018-03-15
Fixes for memory leaks in SPIR-V library.

PawelJurek:
In few places in the SPIR-V library a pattern of “dropAllReferences + removeFromParent” calls was used on values like functions and instructions, that were not meant to be reused. These lead to memory leaks, as these functions do not destruct the objects. The fix is to use “eraseFromParent” instead.
SPIRVDecorate objects were also not deallocated, as they were not added to owning module’s SPIRVEntry list. The fix is to add them, as all other similar entries, e.g. SPIRVDecorationGroup.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVDecorate.h
+++ b/lib/SPIRV/libSPIRV/SPIRVDecorate.h
@@ -105,12 +105,12 @@ protected:
   SPIRVDecorationGroup *Owner; // Owning decorate group
 };
 
-class SPIRVDecorateSet: public std::multiset<const SPIRVDecorateGeneric *,
+class SPIRVDecorateSet: public std::multiset<SPIRVDecorateGeneric *,
     SPIRVDecorateGeneric::Comparator> {
     public:
-  typedef std::multiset<const SPIRVDecorateGeneric *,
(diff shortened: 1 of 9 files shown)


AlexeySotkin:
Could you make commit message 80 symbols wide.

lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -105,12 +105,12 @@ class SPIRVDecorateGeneric:public SPIRVAnnotationGeneric{
   SPIRVDecorationGroup *Owner; // Owning decorate group
 };
 
-class SPIRVDecorateSet: public std::multiset<const SPIRVDecorateGeneric *,
```
AlexeySotkin, line 108: Why did you remove "const" qualifier?

PawelJurek: This is basically because in SPIRVModule.h addEntry function does not take const SPIRVEntries. 
If you take a look at other similar classes, e.g. SPIRVDecorationGroup, they also do not have const qualifiers along the way.

PR #234, opened 2018-03-26
Fix for intel_sub_group_block_read with short type.

PawelJurek:
With previous code, when intel_sub_group_block_read_us scalar function
was being used, we generated __builtin_spirv_intel_sub_group_block_read_p1i16
built-in name. For naming consistency with OpenCL functions, it should have the "us" included, i.e.
__builtin_spirv_intel_sub_group_block_read_us_p1i16


PawelJurek:
@bashbaug @AlexeySotkin Could you please take a look?

AlexeySotkin:
LGTM, Thanks!

PR #238, opened 2019-09-21
Fix compile error: Use explicit conversion of `unique_ptr` to `bool`

mfwitten:
The aging LLVM code is converting  an object of type `unique_ptr` to
an object  of type  `bool`; this is  being done  implicitly, thereby
causing  a compile-time  error  with  a modern,  standards-compliant
compiler.

This is explained well here:

    https://github.com/microsoft/DirectXShaderCompiler/issues/1692

In particular:

> The C++  standard specifies operator  bool on unique_ptr  as being
> explicit,  so this  is  clearly a  bug in  the  LLVM source  code:
> https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_bool

This commit merely makes the  conversion explicit, allowing the code
to be built by GCC version 9.1.0.

This  bug was  fixed  by the  LLVM  project proper  as  part of  the
following commit:

    69341e6abca92f7f118ee7bd99be0cdfc649386f
    2016-04-08
    https://github.com/llvm/llvm-project/commit/69341e6abca92f7f118ee7bd99be0cdfc649386f

### DIFF
--- a/include/llvm/IR/ValueMap.h
+++ b/include/llvm/IR/ValueMap.h
@@ -101,7 +101,7 @@ public:
 
   ~ValueMap() {}
 
-  bool hasMD() const { return MDMap; }
+  bool hasMD() const { return bool(MDMap); }
   MDMapT &MD() {
     if (!MDMap)
       MDMap.reset(new MDMapT);


mfwitten:
Really, the LLVM base for this project should be updated.  
In the meantime, this commit keeps things working.

bader:
@mfwitten, active development of this tool/library has moved to https://github.com/KhronosGroup/SPIRV-LLVM-Translator.
This repo is not maintained anymore, AFAIK.

mfwitten:
Isn't it against the CoC to waste everybody's fscking time? Fix your links/website, you goofballs.

pierremoreau:
@bader Do you know who could answer whether this is still maintained, maybe bring it up on the tooling call?

If it is no longer maintained, the README and/or description should be updated to say so (and point to SPIRV-LLVM-Translator), and the repo archived, imho.

